PR #12, opened 2012-03-09
Retrieve latest version of entities of the same class

merk:
Minor improvement, but I needed a way to retrieve the latest version of each entity of a type so I could display which ones were not 'approved' to the latest available version in my approvals bundle.


smoench:
This needs a rebase against master. Feel free to open this PR again if it is still relevant and rebase has been done.

PR #68, opened 2014-06-24
Fix field name and association

pokap:
Hello,
With the latest version, I found 2 bugs:
- An error of check field name association. The process retrieve the field name of join-column, and adding fields of entity, but the field name of join-column is a column name.
- And an error during hydration, when the association has a join-column without identities.


DavidBadura:
@pokap Can you please create a new PR with tests if the error still exists?

PR #107, opened 2014-12-23
misc bugs with relations

### LINKED ISSUE #87: ManyToOne inheritance bugs
When the ManyToOne mapping to $organisation is on the abstract class I get an exception saying the organisation_id field is not mapped. The _audit table for each subclass includes all other fields from the abstract class except for those that are ManyToOne associations.

Adding the abstract class to audited_entities in config.yml does not stop the error. Moving the ManyToOne mapping for to each subclass using a Trait solves the problem.

Although it's a little extra code moving the mapping to the subclass works for all entities, except one that uses a Doctrine Extensions nested tree - I can't move the tree mappings into the subclasses as it throws an error saying there are multiple keys.

``` yml
#config.yml
simple_things_entity_audit:
    audited_entities:
        - Umm\ProjectBundle\Entity\Project //Note: adding the abstract parent class here does not solve the problem
        //other entities to audit
```

The abstract class that projects inherit from

``` php
/**
* @ORM\Table(name="project_project_abstract")
 * @ORM\Entity(repositoryClass="Umm\ProjectBundle\Repository\AbstractProjectRepository")
 * @ORM\InheritanceType("JOINED")
 * @ORM\DiscriminatorColumn(name="discr", type="string")
 * @ORM\DiscriminatorMap({"project" = "Project", "template" = "ProjectTemplate"})
*/
abstract class AbstractProject extends BaseEntity implements ProjectInterface
{
    /**
     *
     * @ORM\Column(name="title", type="string", length=50)
     * @Assert\Length(
     *  min="3",
     *  max="50",
     *  minMessage="project.title.min",
     *  maxMessage="project.title.max"
     * )
     * @Assert\NotBlank(
     *  message="project.title.not_blank"
     * )
     * @JMS\Groups({"summary", "template", "detail", "timeline", "public_api", "public_api_summary"})
     */
    protected $title; //This property is in the _audit table for each subclass

    /**
     *
     * @ORM\Column(name="description", type="string", length=1000, nullable=true)
     *
     * @Assert\Length(
     *  min="3",
     *  max="1000",
     * )
     * @JMS\Groups({"summary", "template", "detail", "public_api", "public_api_summary"})
     *
     */
    protected $description; //This property is in the _audit table for each subclass

    /**
     * @ORM\ManyToOne(targetEntity="Umm\OrganisationBundle\Entity\Organisation")
     * @ORM\JoinColumn(nullable=true)
     * @JMS\Exclude()
     */
    protected $organisation; //This association is NOT in the _audit table for the subclasses

   //other properties
}
```

The sub class:

``` php
/**
 *
 * @ORM\Entity(repositoryClass="Umm\ProjectBundle\Repository\ProjectRepository")
 * @ORM\Table(name="project_project")
 * @ORM\HasLifecycleCallbacks
 *
 */
class Project extends AbstractProject implements TaskableInterface, ProjectTimelineInterface, LoggableStatisticInterface, CommentableInterface
{
  //Other mappings
}
```

Error description:

"I was stocked to see this PR merged as we have a lot of class inheritance but alas it's not working for us on Symfony 2.6-dev with Sonata Admin. I'm using class table inheritance with Project inheriting from Abstract Project but none of the abstract project properties appear in the project_audit table so I get a DBALException and PDOException when saving or updating a project, or loading fixtures. Not sure if it's caused by Sonata as auditing needs to be set to true in each Admin service?"

Code by @mpclarkson

andrewtch:
Well, misc strange bugs, mostly visible when we load ToOne entitty from not-owning side. Also, fixes edge cases when we can't resolve Abstract Discriminator entity type before we actually load disc value.

Should fix #87 as well. Should be merged only after @mpclarcson checks ).

### DIFF
--- a/src/SimpleThings/EntityAudit/AuditReader.php
+++ b/src/SimpleThings/EntityAudit/AuditReader.php
@@ -208,6 +208,7 @@ class AuditReader
         }
 
         $class = $this->em->getClassMetadata($className);
+
         $tableName = $this->config->getTablePrefix() . $class->table['name'] . $this->config->getTableSuffix();
 
         if (!is_array($id)) {
@@ -236,7 +237,7 @@ class AuditReader
 
             $type = Type::getType($class->fieldMappings[$field]['type']);
             $columnList[] = $tableAlias.'.'.$type->convertToPHPValueSQL(
-                $class->getQuotedColumnName($field, $this->platform), $this->platform) .
+                $class->getColumnName($field, $this->platform), $this->platform) .
                 ' AS ' . $this->em->getConnection()->quote($field);
             $columnMap[$field] = $this->platform->getSQLResultCasing($columnName);
         }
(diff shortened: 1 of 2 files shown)


andrewtch:
@DavidBadura , please merge as checked.

PR #110, opened 2015-01-09
made EA respect extraUpdates

### LINKED ISSUE #9: Using the same entity in both OneToMany and OneToOne
I have found an error while versioning entities like: (error is while auditing Customer entity)

````
class Address
{
    /**
     * @ORM\Column 
     */
    protected $address_text;

    /**
     * @ORM\ManyToOne(targetEntity="Customer", inversedBy="addresses") 
     */
    protected $customer;
}



class Customer 
{
    /**
    * @ORM\OneToMany(targetEntity="Address", mappedBy="customer")
    */
    protected $addresses;

    /**
    * @ORM\OneToOne(targetEntity="Address")
    */
    protected $primary_address;
}
````

the UnitOfWork::getEntityIdentifier throws an notice "Undefined index: ........" 
in UnitOfWork::entityIdentifiers array - the hash is different that the stored one

TESTED ON Doctrine 2.1.3 and 2.1.6

### PR DESCRIPTION
**Pull request by andrewtch**
Closes #9 . Undefined index notices are already fixed in previous commits, but there's a bug when we save tow NEW entities both mapped to each other. The second is is persisted in ```executeExtraUpdates```, now EntityAudit also repeats all extra updates on audit tables.

### DIFF
--- a/src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
+++ b/src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
@@ -23,6 +23,8 @@
 
 namespace SimpleThings\EntityAudit\EventListener;
 
+use Doctrine\ORM\Event\PostFlushEventArgs;
+use Doctrine\ORM\Persisters\BasicEntityPersister;
 use SimpleThings\EntityAudit\AuditManager;
 use Doctrine\Common\EventSubscriber;
 use Doctrine\ORM\Events;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php [hunk 1]
```diff
@@ -81,7 +83,61 @@ public function __construct(AuditManager $auditManager)
...
+        $uowR = new \ReflectionClass($uow);
+        $extraUpdatesR = $uowR->getProperty('extraUpdates');
+        $extraUpdatesR->setAccessible(true);
+
+        $extraUpdates = $extraUpdatesR->getValue($uow);
+
+        foreach ($extraUpdates as $update) {
+            list($entity, $changeset) = $update;
+
+            if ($this->metadataFactory->isAudited(get_class($entity))) {
```
DavidBadura: that's not relevant but early exits are much easy to read.

```php
if (!$this->metadataFactory->isAudited(get_class($entity))) {
    continue;
}
```

thus you havn't deep nesting. the same below.

```php
if (!isset($updateData[$meta->table['name']]) || !$updateData[$meta->table['name']]) {
    continue;
}
```

andrewtch:
Done. It's a known style mistake I'm constantly making )

PR #112, opened 2015-01-21
Proper alias quoting

ShiraNai7:
This PR fixes an issue with aliases:

- all aliases must be properly quoted
    - unquoted aliases get lowercased by PostgreSQL - this prevents `camelCase` properties from being mapped correctly (they remain `NULL`, because the lowercased alias does not match a field name in `AuditReader->createEntity()`)
    - aliases must not be quoted as values by `Connection->quote()`, but as identifiers by `AbstractPlatform->quoteSingleIdentifier()`

I have also removed few useless assignments of `$this->platform` from methods of `AuditReader` (that assignment is already in the constructor, no need to do it again).


andrewtch:
Nice!

Issue #111, opened 2015-01-15
Method prepareUpdateData does not exist 


arlin2050:
Hi,

when I use EntityAudit and SoftDeleteable on the same entity and I remove this entity, I got this error :

ReflectionException: Method prepareUpdateData does not exist in SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php line 114

Seems that entityAudit is launched two times and the second time I got this error. Also, putting this on line 114 resolves the issue but I don't know the consequences : 
``` php
if (!$persisterR->hasMethod('prepareUpdateData')) {
      continue;
}
```

Thanks !

andrewtch:
Hi,

can you give your example entity? Thanks!

arlin2050:
Here is the mapping : 
``` xml
<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping http://doctrine-project.org/schemas/orm/doctrine-mapping.xsd"
        xmlns:gedmo="http://gediminasm.org/schemas/orm/doctrine-extensions-mapping" >
    <entity name="Oru\Bundle\CustomBundle\Entity\Demande" table="oru_custom_demande" repository-class="Oru\Bundle\CustomBundle\Entity\DemandeRepository" >
        <id name="id" type="integer" column="id">
            <generator strategy="AUTO"/>
        </id>
        <field name="statut" type="string" length="20" nullable="true" />
        <field name="deletedAt" column="deleted_at" type="datetime" nullable="true" />
        <gedmo:soft-deleteable field-name="deletedAt" />
    </entity>
</doctrine-mapping>
```

And the EntityAudit configuration : 
```yaml
simple_things_entity_audit:
    audited_entities:
        - Oru\Bundle\CustomBundle\Entity\Demande
```

The php class for this entity is auto-generated via doctrine :
```php
namespace Oru\Bundle\CustomBundle\Entity;

use Doctrine\ORM\Mapping as ORM;

class Custom {

    /**
     * @var integer
     */
    private $id;

     /**
     * @var string
     */
    protected $statut;

    /**
     * @var \DateTime
     */
    private $deletedAt;

    /**
     * Get id
     *
     * @return integer 
     */
    public function getId()
    {
        return $this->id;
    }
    
    /**
     * @param string $statut
     */
[...]

PR #113, opened 2015-01-22
fix for non-scalar fields during extra updates, test case for #111

andrewtch:
- test case showing that #111 indeed works
- fix when we save some non-scalar values during extra updates (e.g. datetime)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .svn
 composer.lock
 vendor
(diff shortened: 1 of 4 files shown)


aivus:
Is it really cache dir should be here?

andrewtch:
Sorry, my bad. Added ignore files.

simplethings guys, can you somehow react on the pr's or it's better if we fork this package and then merge our changes in a bulk sometime later?

Issue #31, opened 2012-10-30
Undefined index when i delete an entity which is audit


psylo31:
Notice: Undefined index: titre in MyPath\vendor\simplethings\entity-audit-bundle\src\SimpleThings\EntityAudit\EventListener\LogRevisionsListener.php line 206

Entity User :

/**
     * @ORM\OneToOne(targetEntity="MyProject\ReveBundle\Entity\Reve", cascade={"persist", "remove"})
     */
    protected $reve;

Action in my Reve repository to remove :

$user->setReve();
$this->_em->persist($user);
$this->_em->remove($reve);
$this->_em->flush();

It's work when the entity is not audit.

Thanks for you help

andrewtch:
@psylo31 , @rickard2 , thsi should be fixed by now. Can you test?

CountZero1981:
Seems that it fixed in latest build. I've had this problem in SonataAdminBundle and it disappeared after update.

andrewtch:
I'm closing this as #117 actually works with no exception. Please comment if you find a way on how to repeat the behavior above.

PR #117, opened 2015-01-30
regression test for issue #31

andrewtch:


OskarStark:
@andrewtch is this already released?

if not, could you create a release?

tolry:
@OskarStark as this was merged in January 2015 (not 2016), this has already been released. Also, these are "just" added tests, no functionality or behavior change was added.

OskarStark:
thanks for the feedback @tolry , didn't get the year correct :)

PR #118, opened 2015-01-31
Add support to separate databases for audit tables

aivus:
Audit tables increase database tables amount twice (when we want audit all tables).
This PR provides support to place audit tables into another database.
This PR fully BC

/cc @malavsky

### DIFF
--- a/src/SimpleThings/EntityAudit/AuditConfiguration.php
+++ b/src/SimpleThings/EntityAudit/AuditConfiguration.php
@@ -34,6 +34,9 @@ class AuditConfiguration
     private $globalIgnoreColumns = array();
     private $currentUsername = '';
     private $revisionIdFieldType = 'integer';
+    private $defaultConnection = 'default';
+    private $revisionConnection = 'default';
+
 
     public function getTablePrefix()
     {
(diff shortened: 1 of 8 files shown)


andrewtch:
I strongly disagree on adding external dependency to a Registry, and introducing Connection1/Connection2/Connection whatever variables. 

Instead, can you refractor this so:
* connection management happens in AuditConfiguration
* additional connection is used only if explicitly set (like configuration-set storage connection etc) 

?

aivus:
@andrewtch, in the current implementation we have only one connection, when "revision" connection information isn't set. When both connections information set - we have 2 EntityManagers:
* First (Default) - for getting information about entity
* Second (Revision) - for audit tables

Registry dependency used when Symfony DI initializes services.
In standalone mode only EntityManager is used.

andrewtch:
And, even more, I am very against have found multiple entity managers.  This is hell. I'm mostly working with EA because of a  application  that had three of them. 

Again-EA can use plain doctrine Connection  for both reading and writing entities (and I did not see you updating Audited Collections for example). Please rewrite this to plain connections and also add tests ensuring that original connection is not being called. 

Спасибо)

DavidBadura:
@aivus I'm also andrews opinion. The versioned tables are not real entities but a kind of copy. So a second entity manager is wrong here. What you want is to save only the versioned data of an entity manager in another database, so you need only a second connection.

I find @andrewtch implementation in pr #120 very good :wink:

andrewtch:
David, thanks for the compliment)) I'll finish that pr this week.

PR #120, opened 2015-02-02
wip implementation of multiple connections

### LINKED ISSUE #64: Store audit data in a separate database
Great extension, thanks!

I'd like to store audit data in a separate database, potentially on a different physical server. 

Is such a thing even possible ? 

If it's just supplying the config option for new entity manager and using it in LogRevisionsListener, then I can do it and create a pull request, it just seems too easy to not be in there already :)

andrewtch:
Closes #64 
Closes #118

### DIFF
--- a/src/SimpleThings/EntityAudit/AuditReader.php
+++ b/src/SimpleThings/EntityAudit/AuditReader.php
@@ -23,6 +23,7 @@
 
 namespace SimpleThings\EntityAudit;
 
+use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\Mapping\ClassMetadata;
@@ -74,6 +75,12 @@ class AuditReader
      * @var bool
      */
     private $loadNativeEntities = true;
+
+    /**
+     * External connection for audit data
+     * @var Connection
+     */
+    private $connection;
 
     /**
      * @return boolean
@@ -149,7 +156,8 @@ class AuditReader
         $this->em = $em;
         $this->config = $config;
         $this->metadataFactory = $factory;
-        $this->platform = $this->em->getConnection()->getDatabasePlatform();
+        $this->connection = $config->getAuditConnection() ?: $this->em->getConnection();
+        $this->platform = $this->getConnection()->getDatabasePlatform();
     }
 
     /**
@@ -157,7 +165,16 @@ class AuditReader
      */
     public function getConnection()
     {
-        return $this->em->getConnection();
+        return $this->connection;
+    }
+
+    /**
+     * @param string $className
+     * @return ClassMetadata
+     */
+    public function getClassMetadata($className)
+    {
+        return $this->em->getClassMetadata($className);
     }
 
     /**
@@ -207,7 +224,7 @@ class AuditReader
             throw new NotAuditedException($className);
         }
 
-        $class = $this->em->getClassMetadata($className);
+        $class = $this->getClassMetadata($className);
 
         $tableName = $this->config->getTablePrefix() . $class->table['name'] . $this->config->getTableSuffix();
 
@@ -261,7 +278,7 @@ class AuditReader
         $joinSql = '';
         if ($class->isInheritanceTypeJoined()
             && $class->name != $class->rootEntityName) {
-            $rootClass = $this->em->getClassMetadata($class->rootEntityName);
+            $rootClass = $this->getClassMetadata($class->rootEntityName);
             $rootTableName = $this->config->getTablePrefix() . $rootClass->table['name'] . $this->config->getTableSuffix();
             $joinSql = "INNER JOIN {$rootTableName} re ON";
             $joinSql .= " re.rev = e.rev";
@@ -284,7 +301,7 @@ class AuditReader
 
         $query = "SELECT " . implode(', ', $columnList) . " FROM " . $tableName . " e " . $joinSql . " WHERE " . $whereSQL . " ORDER BY e.rev DESC";
 
-        $row = $this->em->getConnection()->fetchAssoc($query, $values);
+        $row = $this->getConnection()->fetchAssoc($query, $values);
 
         if (!$row) {
             throw new NoRevisionFoundException($class->name, $id, $revision);
@@ -316,7 +333,7 @@ class AuditReader
      */
     private function createEntity($className, array $data, $revision)
     {
-        $class = $this->em->getClassMetadata($className);
+        $class = $this->getClassMetadata($className);
 
         //lookup revisioned entity cache
         $keyParts = array();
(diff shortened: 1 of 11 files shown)


src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
```diff
@@ -225,7 +225,7 @@ public function postUpdate(LifecycleEventArgs $eventArgs)
     public function onFlush(OnFlushEventArgs $eventArgs)
     {
         $this->em = $eventArgs->getEntityManager();
-        $this->conn = $this->em->getConnection();
+        $this->conn = $this->config->getConnection() ? $this->config->getConnection() : $this->em->getConnection();
```
DavidBadura: the shortcut for

```php
$this->conn = $this->config->getConnection() ? $this->config->getConnection() : $this->em->getConnection();
```

is

```php
$this->conn = $this->config->getConnection() ?: $this->em->getConnection();
```

I think it's also easier to read :wink:

DavidBadura:
:+1: i like it

src/SimpleThings/EntityAudit/AuditReader.php
```diff
@@ -149,7 +156,8 @@ public function __construct(EntityManager $em, AuditConfiguration $config, Metad
         $this->em = $em;
         $this->config = $config;
         $this->metadataFactory = $factory;
-        $this->platform = $this->em->getConnection()->getDatabasePlatform();
+        $this->connection = $config->getConnection() ? $config->getConnection() : $this->em->getConnection();
```
aivus: `$config->getConnection() ?: $this->em->getConnection();`

andrewtch:
OK, Im always afraid of short terniary somehow)  some more tests, especially with collections are needed though.

aivus:
:+1: thanks for this implementation

aivus:
Don't forget to change `auditable.xml` and DI configuration, please. Thanks

andrewtch:
That's why it's still WIP. I'm porting tests currently.

andrewtch:
Ouch, hopefully finished.

@aivus , can you please test it in your environment?
@DavidBadura , can you check the code, especially at https://github.com/simplethings/EntityAudit/pull/120/files#diff-526bd1905ced8959412a89500653b48aR62 - is it a correct approach or a sh**code? Because else we'd need to use container expressions (like @doctrine.getConnection(audit)) or introduce whole Repository, which would make problems for no-symfony projects.

DavidBadura:
tonight i will send a pull request :wink:

andrewtch:
ok, i'll wait then (though you can just offer a better approach if you can explain it)

andrewtch:
@aivus , can you pelase test this?

andrewtch:
@aivus , ping!

aivus:
I will test it on this week

src/SimpleThings/EntityAudit/DependencyInjection/SimpleThingsEntityAuditExtension.php
```diff
@@ -58,5 +56,13 @@ public function load(array $configs, ContainerBuilder $container)
         if (true === $config['listener']['current_username']) {
             $loader->load('current_username.xml');
         }
+
+        if ($config['audit_connection']) {
+            $container->getDefinition('implethings_entityaudit.config')
```
aivus: It should be "simplethings_entityaudit.config"

aivus:
I have `revision` connection, but when I try set
`audit_connection: revision` in `simple_things_entity_audit` section in config.yml i got:
`The service "simplethings_entityaudit.config" has a dependency on a non-existent service "revision".`

How can I set connection for `audit_connection`?

src/SimpleThings/EntityAudit/DependencyInjection/SimpleThingsEntityAuditExtension.php
```diff
@@ -58,5 +56,13 @@ public function load(array $configs, ContainerBuilder $container)
...
+                ->addMethodCall('setAuditConnection', array(
+                        new Reference($config['audit_connection'])
```
aivus: Hm, it works when I use:
`new Reference('doctrine.dbal.' . $config['audit_connection'] . '_connection')`

aivus:
When I use `app/console doctrine:schema:update --dump-sql` I expect get sql, but get created audit tables in database.

andrewtch:
Yes, it's not a bug, it was intended to use doctrine.dbal.CONNECTIONNAMEHERE_connection as a parameter.

However, there is indeed a big hole in schema generation. @DavidBadura , can you take a look on issue below?

The thing is that we need to create audit tables for a different connections; as you can see in ```CreateSchemaListener``` we need to determine which connection we're using.

For ```postGenerateSchema``` it's a simple as ```$entArgs->em->getConnection()``` - this issue is that the connection is not accessible in ```postGenerateSchemaTable```. Do you know a way to get a connection instance in ```postGenerateSchemaTable``` so we can do something like:

```php
if ($this->config->getAuditConnection() && $em->getConnection() != $this->config->getAuditConnection()) {
    //skip tables for uknown connections

    return;
}
```

?

andrewtch:
I've made a very basic implementation of multiple connections that cleans up uneeded tables for connections that are not audit; however, we still need to add support of multiple connections on service definition level.

@DanielBadura , can you review?

DavidBadura:
I have no time this week. I will review at the weekend. :-/

andrewtch:
David? )

andrewtch:
@DavidBadura , ping?

DavidBadura:
Hey @andrewtch it looks very good. What do you mean with "we still need to add support of multiple connections on service definition level."?

andrewtch:
Em, you see...

In postGenerateSchemaTable I create audit tables. These are created for the main connection, not for the audited. 

Yes, there's a workaround in postGenerateSchema that skips all "normal" tables from creation, but this is way ugly since we force the entities to be bound to audit connection (thus having 2x EntityManagers).

 So the question is how to correctly schedule table creation for one connection when actually creating entities for another?

Filoz:
Great job @andrewtch !
will it be merged?

andrewtch:
Closing as obsolete

bendavies:
@andrewtch was the only problem with this that the SchemaListener doesn't work?

andrewtch:
@bendavies , yes, but it whould require majot workaround to deal with this.

bendavies:
@andrewtch yeah, that's what i thought. did you have any ideas? i couldn't see anything quickly...

andrewtch:
I also can't except for some reflection woodoo and messing up schemalistener (see https://github.com/simplethings/EntityAudit/pull/120#issuecomment-95834844 and https://github.com/simplethings/EntityAudit/pull/120#issuecomment-83614356)

bendavies:
Looks like to me we can't use a schema listener if it's on it's own new connection.
I would say we need to use a command to list the sql needed to create the diff much like how `init:acl` works in the symfony security bundle.

PR #126, opened 2015-03-31
require DBAL 2.5

mernen:
#119 (commit 912303e7) uses `AbstractPlatform::getIdentitySequenceName`,
which is a new method introduced in DBAL 2.5 (see doctrine/dbal@6b4ba50f).


DavidBadura:
thank you :wink:

PR #127, opened 2015-04-21
Generate database schema when class is extends with abstract class

jbdelhommeau:
Currently when class is extended with abstract class the database schema is not generated with doctrine:schema:update. This commit fix that.

### DIFF
--- a/src/SimpleThings/EntityAudit/EventListener/CreateSchemaListener.php
+++ b/src/SimpleThings/EntityAudit/EventListener/CreateSchemaListener.php
@@ -60,7 +60,7 @@ class CreateSchemaListener implements EventSubscriber
     {
         $cm = $eventArgs->getClassMetadata();
 
-        if (!$this->metadataFactory->isAudited($cm->name)) {
+        if (!$this->metadataFactory->isClassAudited($cm)) {
             $audited = false;
             if ($cm->isInheritanceTypeJoined() && $cm->rootEntityName == $cm->name) {
                 foreach ($cm->subClasses as $subClass) {
--- a/src/SimpleThings/EntityAudit/Metadata/MetadataFactory.php
+++ b/src/SimpleThings/EntityAudit/Metadata/MetadataFactory.php
@@ -32,11 +32,27 @@ class MetadataFactory
         $this->auditedEntities = array_flip($auditedEntities);
     }
 
-    public function isAudited($entity)
+    public function isAudited($entityName)
     {
-        return isset($this->auditedEntities[$entity]);
+        return isset($this->auditedEntities[$entityName]);
     }
-    
+
+    public function isClassAudited($entityClass)
+    {
+        if ($this->isAudited($entityClass->name)) {
+            return true;
+        }
+        if ($subClasses = $entityClass->subClasses) {
+            foreach ($subClasses as $oneEntityclass) {
+                if ($this->isAudited($oneEntityclass)) {
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
+
     public function getAllClassNames()
     {
         return array_flip($this->auditedEntities);


andrewtch:
Looks fine for me (at least it will allow not to specify all the classes in hierarchy), but pretty please do fix unit test for newer versions (see travis).

jbdelhommeau:
It's work with PHPUnit 4.0.0 but not with PHPUnit 4.6.4 I don't know why :(

andrewtch:
Can you take a look on the build? https://travis-ci.org/simplethings/EntityAudit/jobs/59362013 ? It's not a phpunit issues, it just breaks things with 1:M somehow...

jbdelhommeau:
Yes I looked. But even if I try without my change the unit test don't pass...
It's not with my little change :( I don't know why

sebastianreloaded:
Well, currently the master branch tests will fail.

There are several errors of this kind:
Entity 'SimpleThings\EntityAudit\Tests\AnimalAudit' has to be part of the discriminator map of 'SimpleThings\EntityAudit\Tests\AnimalAudit' to be properly mapped in the inheritance hierarchy. Alternatively you can make 'SimpleThings\EntityAudit\Tests\AnimalAudit' an abstract class to avoid this exception from occurring.

Also 2 errors with already existing tables:
SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'escapedcolumnsentity' already exists' while executing DDL: CREATE TABLE EscapedColumnsEntity (id INT AUTO_INCREMENT NOT NULL, lft INT NOT NULL, `left` INT NOT NULL, PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB

and more

jbdelhommeau:
It work's !
Unit test suite don't work with doctrine/orm >= 2.5.
I fixed version on composer.json

sebastianreloaded:
Hmm this conflicts with the already merged https://github.com/simplethings/EntityAudit/pull/126

sebastianreloaded:
true, doctrine/orm 2.5 was released 21 days ago

jbdelhommeau:
doctrine/orm ~2.2 update all version before branch 3

2015-04-23 15:31 GMT+02:00 sebastianreloaded <notifications@github.com>:

sebastianreloaded:
jep, got that ;)

andrewtch:
Guys, it's cool, but we have to support backwards compatibility. I can not accept PR that rises doctrine to 2.4 when symfony is bundled against ~2.2. Can you find another fix?

DavidBadura:
We can't force the version to 2.4 because we would save backwards compatibility in feature releases. We need a different approach.

andrewtch:
..and future compatibility too, so please remove <2.5 and stick to ~2.2.

jbdelhommeau:
Yes but your unit test suite don't work with actual last version 2.5...

2015-04-24 9:23 UTC+02:00, Andrew Tch <notifications@github.com>:

andrewtch:
Then we have to fix it, because symfony works fine with 2.5 afaik.

DavidBadura:
i have restart travis for current master: https://travis-ci.org/simplethings/EntityAudit/builds/56563337
so we can see what is broken

sebastianreloaded:
But the last pr in master already breaks symfony compatibility, which has doctrine/dbal < 2.5. Then this should be reverted too.

sebastianreloaded:
they also removed doctrine/orm 2.5 support: https://github.com/symfony/symfony-standard/commit/acd2c335957a0bc2377367c0547a50f4eb05f1a2

DavidBadura:
Please create a new PR if the error still exists, thanks!

PR #129, opened 2015-04-23
Support for getting a list of all revisions that were made of a specific entity class

sebastianreloaded:
Currently it is not possible to search the revision history for specific audited entities. This will allow you to add a filter with one or more entity classes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
     "description": "Audit for Doctrine Entities",
     "require": {
         "doctrine/dbal": "~2.5",
-        "doctrine/orm": "~2.2"
+        "doctrine/orm": "2.4.*"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


andrewtch:
@sebastianreloaded , can you:
* squash the commits
* use ~2.2 as version constraint? probably some more fixes may apply, but I don't want to raise the requirement
* update the doc?

Thanks!

sebastianreloaded:
@andrewtch no problem, will do that.

SirWaddles:
Hey,

Just working on something lately and this would be a pretty useful feature.

Alternative seems to be to get the history of each entity in a collection which ends up being over 1000 database queries in my test run.

DavidBadura:
Hey, can you revert the changes in the composer.json and rebase it? Thanks!

PR #130, opened 2015-04-23
Support foreign key changes in diff

sebastianreloaded:
Currently foreign key changes are not considered by diff. This will return those changes too.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
     "description": "Audit for Doctrine Entities",
     "require": {
         "doctrine/dbal": "~2.5",
-        "doctrine/orm": "~2.2"
+        "doctrine/orm": "2.4.*"
     },
     "require-dev": {
         "symfony/framework-bundle": "~2.3",
--- a/src/SimpleThings/EntityAudit/AuditReader.php
+++ b/src/SimpleThings/EntityAudit/AuditReader.php
@@ -748,10 +748,22 @@ class AuditReader
     {
         $metadata = $this->em->getClassMetadata($className);
         $fields = $metadata->getFieldNames();
+        $associationsFields = $metadata->getAssociationNames();
 
         $return = array();
         foreach ($fields AS $fieldName) {
             $return[$fieldName] = $metadata->getFieldValue($entity, $fieldName);
+        }
+
+        foreach($associationsFields as $key => $association) {
+            $associationValues = $entity->{'get' . ucfirst($association)}();
+            $values = [];
+            if (get_class($associationValues) == 'Doctrine\Common\Collections\ArrayCollection') {
+                $values[] = count($associationValues) > 0 ? array_map([$associationValues->first(), 'getId'], $associationValues) : [];
+            } else {
+                $values[] = $associationValues == null ? null : $associationValues->getId();
+            }
+            $return[$association] = $values;
         }
 
         return $return;
--- a/src/SimpleThings/EntityAudit/AuditReader.php
+++ b/src/SimpleThings/EntityAudit/AuditReader.php
@@ -758,8 +758,8 @@ class AuditReader
         foreach($associationsFields as $key => $association) {
             $associationValues = $entity->{'get' . ucfirst($association)}();
             $values = array();
-            if (get_class($associationValues) == 'Doctrine\Common\Collections\ArrayCollection') {
-                $values[] = count($associationValues) > 0 ? array_map(array($associationValues->first(), 'getId'), $associationValues) : array();
+            if ($associationValues instanceof \Doctrine\Common\Collections\Collection) {
+                $values[] = count($associationValues) > 0 ? array_map(array($associationValues->first(), 'getId'), $associationValues->toArray()) : array();
             } else {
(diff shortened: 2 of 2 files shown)


sebastianreloaded:
relates to https://github.com/simplethings/EntityAudit/issues/123

eigan:
Why has this not been merged? Is there any problem with this approach? @tolry?

eigan:
Sorry, should have tested. This doesn't really work that well.

DavidBadura:
Hey, can you revert the changes in the composer.json and create a test? Thanks!

composer.json
```diff
@@ -5,7 +5,7 @@
     "description": "Audit for Doctrine Entities",
     "require": {
         "doctrine/dbal": "~2.5",
-        "doctrine/orm": "~2.2"
+        "doctrine/orm": "2.4.*"
```
Soullivaneuh: Why changing that and BTW why `2.4.*`? It should be `^2.4`.

src/SimpleThings/EntityAudit/AuditReader.php
```diff
@@ -748,12 +748,24 @@ public function getEntityValues($className, $entity)
...
         $fields = $metadata->getFieldNames();
+        $associationsFields = $metadata->getAssociationNames();
 
         $return = array();
         foreach ($fields AS $fieldName) {
             $return[$fieldName] = $metadata->getFieldValue($entity, $fieldName);
         }
 
+        foreach($associationsFields as $key => $association) {
+            $associationValues = $entity->{'get' . ucfirst($association)}();
```
Soullivaneuh: This is not reliable.

A getter for `available` property can be `isAvailable` for example.

You should use the [Symfony PropertyAccess component](https://symfony.com/doc/current/components/property_access.html).

src/SimpleThings/EntityAudit/AuditReader.php
```diff
@@ -748,12 +748,24 @@ public function getEntityValues($className, $entity)
...
+            $values = array();
+            if ($associationValues instanceof \Doctrine\Common\Collections\Collection) {
```
Soullivaneuh: The FQCN should be imported.

src/SimpleThings/EntityAudit/AuditReader.php
```diff
@@ -748,12 +748,24 @@ public function getEntityValues($className, $entity)
...
+                $values[] = count($associationValues) > 0 ? array_map(array($associationValues->first(), 'getId'), $associationValues->toArray()) : array();
```
Soullivaneuh: * `count($associationValues) > 0` can be simplified with `count($associationValues)`
* This line is way too long. You can break it on ternary steps
* Some comments of why you are doing that would be great IMHO, it's hard to understand, so to maintain.

Soullivaneuh:
@eigan @DavidBadura I also need this feature for my project.

BTW, it seems @sebastianreloaded  try to compare all properties of the foreign association. But we just need the reference diff to see if the link is the same, right?

Plus, this PR is very old, not sure it will be updated.

I'll try to work for a new one.

src/SimpleThings/EntityAudit/AuditReader.php
DavidBadura: We shouldn't use getters. You don't have it for each property (for example if you develop DDD). We should diff over properties and use `$metadata->getFieldValue($entity, $fieldName)` for it.

DavidBadura:
@Soullivaneuh with pleasure :)

Soullivaneuh:
@DavidBadura On which branch should I work?

For me it's for a minor release because it's more a new feature than a bugfix.

Branch 1.0 is for 1.0.x releases if I'm right but master is already pointing for 2.x.

Can you please tell me what to do?

Maybe create a 1.x branch instead of 1.0?

src/SimpleThings/EntityAudit/AuditReader.php
Soullivaneuh: Didn't know about `$metadata->getFieldValue`. It's the same behavior as property access but for doctrine. Sounds the way to go.

DavidBadura:
We do not want any new features anymore for 1.0. So we do not have feature releases for version 1.

The master will be version 2.0. We haven't ressources to support two versions, so we will no longer support branch 1.0 as version 2.0 was released.

So please every feature in the master :)

src/SimpleThings/EntityAudit/AuditReader.php
DavidBadura: No, `property access` use getter methods to access the values. Doctrine doesn't use class methods. It uses class/property reflections to access the properties directly.

Soullivaneuh:
Ok. I would recommend to use 1.x, 2.x etc branch and master for next major.

2.x will be the maintained branch for both patch and minor, master for major only.

It's easier to manage and keep some flexibility with semver. I use this for my personal projects with not a lot of time for them, works great. :wink:

Soullivaneuh:
@DavidBadura Please see #249.

DavidBadura:
replaced by #249

PR #134, opened 2015-06-04
2.2 deprecation fixes

Rvanlaak:
The `pattern` option is deprecated since 2.2, and need to be replaced with `path`.


Rvanlaak:
@DavidBadura a new release would be awesome! :)

DavidBadura:
i have to check before why the build is broken :-/

Issue #149, opened 2015-10-08
Wrong Type Hint for EntityManager 


igst:
Hi!

If a custom EntityManager Decorator is used, AuditManager->createAuditReader() and AuditReader::__construct() throwing exceptions because of 'EntityManager' Type Hint: 

*Catchable Fatal Error: Argument 1 passed to SimpleThings\EntityAudit\AuditReader::__construct() must be an instance of Doctrine\ORM\EntityManager, instance of ACME\Bundle\SomeBundle\Doctrine\EntityManagerDecorator given*

and 

*Catchable Fatal Error: Argument 1 passed to SimpleThings\EntityAudit\AuditManager::createAuditReader() must be an instance of Doctrine\ORM\EntityManager, instance of ACME\Bundle\SomeBundle\Doctrine\EntityManagerDecorator given*

I think it would be better to use 'EntityManagerInterface' instead. 

Thanks!

tolry:
Sounds about right - would you mind creating a PR?

TomasVotruba:
This can be closed now

DavidBadura:
thank you :)

OskarStark:
:+1:

PR #152, opened 2015-10-09
AuditReader: Use interface typehint for entityManager

TomasVotruba:
Ref #149

### DIFF
--- a/src/SimpleThings/EntityAudit/AuditReader.php
+++ b/src/SimpleThings/EntityAudit/AuditReader.php
@@ -25,6 +25,7 @@ namespace SimpleThings\EntityAudit;
 
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\EntityManager;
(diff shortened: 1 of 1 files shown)


DavidBadura:
thank you :-)

TomasVotruba:
Thanks!

src/SimpleThings/EntityAudit/AuditReader.php
```diff
@@ -25,6 +25,7 @@
 
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\EntityManager;
```
OskarStark: is this already used in this class?

TomasVotruba: @OskarStark Good point. #154

OskarStark: :smile:

PR #155, opened 2015-10-16
Supporting ORM 2.5

smoench:


OskarStark:
The Indention fix of the composer.json should ne in a separate PR to make only the version Change part of this PR

Rvanlaak:
:+1: if this could be merged before symfony `2.8` is released

composer.json
```diff
@@ -1,19 +1,32 @@
...
+        "Persistence",
+        "Database",
+        "Audit"
+    ],
     "description": "Audit for Doctrine Entities",
     "require": {
         "doctrine/dbal": "~2.5",
-        "doctrine/orm": "~2.2"
+        "doctrine/orm": "~2.2",
+        "doctrine/doctrine-bundle": "~1.2"
```
DavidBadura: why is it required?

rvanlaak:
ping @DavidBadura

composer.json
rvanlaak: think this can be `~1.4`

Soullivaneuh:
@smoench What is missing for get this merged? Need some help?

Is this PR working actually?

smoench:
@Soullivaneuh one test is failing https://travis-ci.org/simplethings/EntityAudit/jobs/88157908#L294 
the failure was reported in #31 but the test case seems working in master (orm 2.4)

I think LogRevisionsListener::onFlush() and LogRevisionsListener::postFlush() needs some more improvements.

Soullivaneuh:
Problem is related to issue deletion: https://github.com/simplethings/EntityAudit/blob/26134f53b4725a7a14e4e6df365943d7bcc4c390/tests/SimpleThings/Tests/EntityAudit/IssueTest.php#L101

If I comment it, test passes.

rvanlaak:
Shouldn't you refresh the entity before removing it?

```php
$this->em->refresh($reve);
```

Hope that solves it

Soullivaneuh:
@rvanlaak Nope, this change nothing.

smoench:
@rvanlaak @Soullivaneuh tests are running green. would you please so kind as to test this branch in your projects. thanks in advance!

rvanlaak:
Great! Will do that before the end of the year
Op 29 dec. 2015 6:59 p.m. schreef "Simon Mönch" <notifications@github.com>:

phpunit.xml.dist
```diff
@@ -23,5 +23,6 @@
         <var name="DOCTRINE_USER" value=""/>
         <var name="DOCTRINE_PASSWORD" value=""/>
         <var name="DOCTRINE_HOST" value=""/>
+        <!--<var name="DOCTRINE_PATH" value="./db.sqlite"/>-->
```
rvanlaak: This can be removed?

rvanlaak:
I've tested the branch, and all my tests still pass. Added two minor comments :+1: 

Would be great if you could introduce a `CHANGELOG.md` with the new features?

composer.json
```diff
@@ -1,19 +1,38 @@
...
     "description": "Audit for Doctrine Entities",
     "require": {
+        "php": "^5.3.9",
         "doctrine/dbal": "~2.5",
-        "doctrine/orm": "~2.2,<2.5"
+        "doctrine/orm": "~2.4"
     },
     "require-dev": {
+        "doctrine/doctrine-bundle": "~1.4",
         "symfony/framework-bundle": "~2.3",
```
rvanlaak: Can we make `~2.3|~3.0` of this? Or maybe even `~2.7|~3.0`?

phpunit.xml.dist
smoench: No, this is great for debugging ...

composer.json
smoench: I will create an PR in the next few days as some deprecated features are still be used.

smoench: Btw this is only a `require-dev` but it wouldn't working with symfony `~3.0` as of #165

src/SimpleThings/EntityAudit/AuditConfiguration.php
```diff
@@ -35,6 +37,23 @@ class AuditConfiguration
...
+    public function getTableName(ClassMetadataInfo $metadata)
+    {
+        $tableName = $metadata->getTableName();
+
+        //## Fix for doctrine/orm >= 2.5
+        if (method_exists($metadata, 'getSchemaName') && $metadata->getSchemaName()) {
+            $tableName = $metadata->getSchemaName() . '.' . $tableName;
+        }
+
+        return $this->getTablePrefix() . $tableName . $this->getTableSuffix();
```
tolry: Doesn't this concatination need to be done before prefixing the schema name? Else you might end up having `$tablePrefix.$schemaName.$tableName.$tableSuffix`

composer.json
```diff
@@ -1,19 +1,38 @@
 {
     "name": "simplethings/entity-audit-bundle",
-    "license": "New BSD",
-    "keywords": ["Persistence", "Database", "Audit"],
+    "license": "LGPL-2.1",
```
tolry: Why the license change?

smoench: The license file says it is `LGPL-2.1`

src/SimpleThings/EntityAudit/AuditConfiguration.php
smoench: Indeed but with ORM 2.4 you would have the same result.

smoench: opened an issue #166

OskarStark:
:+1:

Issue #83, opened 2014-11-22
Audit Events


andrewtch:
It would be nice to have internal events for audit manger (like "before audit write") etc, that could modify audited data (like converting 1234 5678 0000 0000 card number to 1234 ***\* ***\* **00).

If this behavior is OK, I'll make a PR.

OskarStark:
great idea!

PR #170, opened 2016-01-22
Change LogRevisionsListener methods to protected

rvanlaarhoven:
Changed all private methods in `LogRevisionsListener` to protected in order to be able to extend the listener without having to override everything.

### DIFF
--- a/src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
+++ b/src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
@@ -41,47 +41,47 @@ class LogRevisionsListener implements EventSubscriber
     /**
      * @var \SimpleThings\EntityAudit\AuditConfiguration
      */
-    private $config;
+    protected $config;
 
     /**
      * @var \SimpleThings\EntityAudit\Metadata\MetadataFactory
      */
-    private $metadataFactory;
+    protected $metadataFactory;
 
     /**
      * @var \Doctrine\DBAL\Connection
      */
-    private $conn;
+    protected $conn;
 
     /**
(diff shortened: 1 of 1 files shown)


OskarStark:
seem reasonable to me

DavidBadura:
change everything to protected is dangerous. its more difficult to changes something without bc-breaks for us. what do you want override?

smoench:
:-1:

tolry:
As @DavidBadura said, making properties/methods protected makes that part of your code effectively public API when dealing with libraries that are extended by developers. Making it necessary to consider BC for every possible refactoring in this code.

@rvanlaarhoven it would be really helpful to know, what you intend to extend/modify. Mostly extension by composition is less error prone, than by inheritance. Probably events might be a better solution here than extending the class.

rvanlaarhoven:
Actually, we decided to stick with our old solution (Gedmo Loggable) for now, until we have some more time to refactor. so this PR is not directly necessary.

To clarify what I intended to do is override `saveRevisionEntityData`, to;
- Add another column to the audit table with extra log data.
- We're using versioning on a nested entity structure, with publish functionality. Not every change needed a new audit row, so when a log was already created and the version wasn't published yet, I needed to execute an UPDATE instead of an INSERT.

tolry:
@rvanlaarhoven thx for clarification, OK if I close this PR for now? If this gets relevant for you later, I would like to work on a different solution (e.g. events, see above) for this. Providing special hooks for extending the existing functionality.

DavidBadura:
ref #83 ??

rvanlaarhoven:
@tolry Thanks!

PR #171, opened 2016-01-23
Allow usage on php7

rvanlaarhoven:
Updated the composer php requirements to allow the usage on any php version greater than or equal to `5.3.9`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7.0
   - hhvm
   
 before_script:
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
     ],
     "description": "Audit for Doctrine Entities",
     "require": {
-        "php": "^5.3.9",
+        "php": ">=5.3.9",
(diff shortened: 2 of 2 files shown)


tolry:
Did anyone test this on php7? :smile: Could you add php7 to the travis configuration? Then we could see at least if the tests pass.

rvanlaarhoven:
That makes sense :). I am using this on php 7 locally and haven't had any problems yet.

OskarStark:
:+1:

DavidBadura:
:+1:

tolry:
@rvanlaarhoven thx for contributing and testing this

smoench:
@rvanlaarhoven greater than version constraints are really bad as they are allow any new major version and you don't want to support PHP 6 or 8 yet :wink:

tolry:
@smoench remembered it this way too, but checked with symfony/symfony before merging - at least they are using the same constraint, see https://github.com/symfony/symfony/blob/master/composer.json#L19

Feel free to create a PR containing something like '~5.x|7.0.*' though.

DavidBadura:
He has already changed :wink:

PR #162, opened 2015-11-25
Find revisions by date and/or username

mathieu-axiocode:
Hi, 
If you think searching revisions by date range or username could be usefull, I have just added a method to your AuditReader to manage this kind of request.

PR #177, opened 2016-02-19
add the possibility to find by date

thomasbeaujean:
Hello,

Use case:
I have an entity that is tracked by your bundle.
I would like to get the entity by the datetime instead of the version.

So I added this method to have directly the entity at the date given.

### DIFF
--- a/src/SimpleThings/EntityAudit/AuditReader.php
+++ b/src/SimpleThings/EntityAudit/AuditReader.php
@@ -848,4 +848,61 @@ class AuditReader
 
         return $result;
     }
+
+    /**
+     * Find a class at the specific date.
+     *
+     * This method does not require the date to be exact but it also searches for an earlier revision
+     * of this entity and always returns the latest revision below or equal the given date
+     *
+     * @param string   $className
+     * @param mixed    $id
+     * @param DateTime $date
+     *
+     * @return object
+     *
+     * @throws AuditException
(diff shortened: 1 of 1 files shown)


src/SimpleThings/EntityAudit/AuditReader.php
```diff
@@ -848,4 +848,61 @@ public function getEntityHistory($className, $id)
...
+     *
+     * @param \DateTime $date
+     *
+     * @return int|null The revision
+     */
+    public function getRevisionByDate(\DateTime $date)
+    {
+        $revision = null;
+
+        $query = "SELECT revisions.id FROM ".$this->config->getRevisionTableName()." WHERE timestamp <= '".$date->format("Y-m-d H:i:s")."' ORDER BY revisions.timestamp DESC,revisions.id DESC LIMIT 1";
```
Soullivaneuh: The method name does not match what it does IMHO.

You fetch all revisions **before** the given date, but looking at the name of the method, it looks like it should return revision on a **specific date**.

BTW, why not accept a begin and end datetime arguments?

DavidBadura:
Thank you for your work! But I think a criteria object is the better solution: #162

thomasbeaujean:
@DavidBadura You are right, the criteria will be more usefull, I will wait for the #162 to update my app.

PR #178, opened 2016-03-23
Grab 'name' key from joinColumn in association mapping for column name.

jamescdavis:
Each element in the joinColumns array in an association mapping is an associative array. When pulling out the column name, it needs to access array key 'name'.


tolry:
Hi @jamescdavis - thanks for your contribution!

If I understand correctly this currently results in an error or empty column name? Could you write a regression test reproducing this problem?

jamescdavis:
I have discovered there are more changes needed to get auditing of entities identified through foreign entities to work, so I am closing this pull request and I will open another (which will include a regression test). The changes in this pull request had worked only because the foreign entity I was testing on happened to have a __toString() that returned the id. My next pull request will include a complete fix.

PR #179, opened 2016-03-25
Fix: the entity must be search by the ids instead of the identifiers

thomasbeaujean:
The entities must be search by the ids and not the identifiers.
The id array contains the column_name and not the field_name.
So the whereSql have to use directly the column_name given.

The issue:
There is an audtited entity A that have an OneToOne audited entity B
The B entity has the owningSide.

When the A is loaded, it tries to load the B but an Error is thrown.

The id parameter of the find function contains (a_id => XXX) but the whereSql is set with the string "and id = XXX" instead of "a_id = XXX"


DavidBadura:
The tests failed with your change. If the problem still exists, please create a new PR with a unit test. Thank you!

PR #180, opened 2016-04-05
Update all fields in one go

loostro:
This change updates all audited fields in one query, drastically reduceing the amount of queries.

It is still unclear to me, why would the author choose to execute 1  UPDATE query per each field?
And why UPDATE audit table with the same data (parameters) that were INSERTED?

To me this (these) UPDATE (s) are completely redundant.
### Previously

For audited table `transaction` with fields `type` and `value` this bundle executed:
- START TRANSACTION
- INSERT INTO `transaction` (value, type) VALUES ('foo', 'bar')
- INSERT INTO `revisions`(timestamp, username) VALUES ('2016-04-05 14:07:28', 'anon.')
- INSERT INTO `transaction_audit` (rev, revtype, id, value, type) VALUES ('1', 'INS', 1, 'foo', 'bar')
- COMMIT
- UPDATE `transaction_audit` SET `value` = 'foo' WHERE rev = '1' AND id = 1
- UPDATE `transaction_audit` SET `type` = 'bar' WHERE rev = '1' AND id = 1
- ... if table has 20 fields, here will be 20 update queries ...
### After

For audited table `transaction` with fields `type` and `value` this bundle executed:
- START TRANSACTION
- INSERT INTO `transaction` (value, type) VALUES ('foo', 'bar')
- INSERT INTO `revisions`(timestamp, username) VALUES ('2016-04-05 14:07:28', 'anon.')
- INSERT INTO `transaction_audit` (rev, revtype, id, value, type) VALUES ('1', 'INS', 1, 'foo', 'bar')
- COMMIT
- UPDATE `transaction_audit` SET `value` = 'foo', `type` = 'bar' WHERE rev = '1' AND id = 1
- ... if table has 20 fields, they all will be updated in one query above ...
### Question

Why INSERT then UPDATE the same parameters?

### DIFF
--- a/src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
+++ b/src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
@@ -110,57 +110,66 @@ class LogRevisionsListener implements EventSubscriber
                 continue;
             }
 
+            $params = array();
+            $sql = 'UPDATE ' . $this->config->getTableName($meta) . ' ';
+
+            $firstField = true;
             foreach ($updateData[$meta->table['name']] as $field => $value) {
-                $sql = 'UPDATE ' . $this->config->getTableName($meta) . ' ' .
-                    'SET ' . $field . ' = ? ' .
-                    'WHERE ' . $this->config->getRevisionFieldName() . ' = ? ';
-
-                $params = array($value, $this->getRevisionId());
-
-                $types = array();
-
-                if (in_array($field, $meta->columnNames)) {
-                    $types[] = $meta->fieldMappings[$meta->getFieldForColumn($field)]['type'];
+                if ($firstField) {
+                    $sql .= "SET `$field` = ? ";
+                    $firstField = false;
                 } else {
-                    //try to find column in association mappings
-                    $type = null;
+                    $sql .= ", `$field` = ? ";
+                }
+                $params[] = $value;
+            }
 
-                    foreach ($meta->associationMappings as $mapping) {
-                        if (isset($mapping['joinColumns'])) {
-                            foreach ($mapping['joinColumns'] as $definition) {
-                                if ($definition['name'] == $field) {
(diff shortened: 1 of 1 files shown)


tolry:
@loostro Sounds like a great improvement, thanks for your contribution. There seems to be a minor test failure, could you take a look at that?

Since I am no expert for this part of the codebase

* is there anyone still around, who has first hand experience with this?
* are there enough tests in place, to make sure this doesn't break anything?

loostro:
@tolry I have created a small app as a proof of concept (simple insert/edit auditioned entity app). In my app I have completely removed the onFlush event (no updates at all). That seems like an even better approach as:

| test |  master | this PR  | my APP  |
|---|---|---|---|
| new entity with 5 fields | 1 insert, 5 updates  | 1 insert, 1 update  | 1 insert, 0 updates  |
| edit entity with 5 fields  | 1 insert, 5 updates  | 1 insert, 1 update  | 1 insert, 0 updates  |

so it seems like these updates are completely useless... but there might be something that I'm missing here? I'm really interested to hear why this was introduced, is there some case I'm not considering?

PS. I will publish test app soon and link here.

ravindrakhokharia:
@loostro @tolry 
Any news yet on this issue ? 
It seems its increasing performance when we have large set of data !

loostro:
@ravindrakhokharia we've overriden the eventListener removeing the useless updates and its in production since Apr 2016

tolry:
@loostro there were larger changes to the test suite, could you please rebase your branch and check if the unit test failure still remains?

@bendavies @andrewtch any thoughts about this changes? Any dangers for data consistencies, edge cases not covered, ...?

bendavies:
This seems like a large perf boost. would like to get this in,

bendavies:
@loostro can you resolve conflicts so this an be merged?

DavidBadura:
closed it in favor of #201

PR #186, opened 2016-06-02
Allow to work when the associations don't have the revision

### LINKED ISSUE #157: OneToOne associations not supported when owning side does not exist
Consider the following entities:

```
class Human
{
    /**
     * @ORM\OneToOne(targetEntity="Robot", mappedBy="human")
     */
    property $robot;
}

class Robot
{
    /**
     * @ORM\OneToOne(targetEntity="Human", inversedBy="robot")
     */
    property $human;
}
```

If you persist a Human without a Robot, attempting to load an instance of that Human from AuditReader::find() will fail, as it attempts to find a Robot with the corresponding `human_id`. It currently seems to assume that both sides of a OneToOne association must exist. This is the workaround I am using:

```
try {
    $value = $this->find($targetClass->name, $pk, $revision, array('threatDeletionsAsExceptions' => true));
} catch (DeletedException $e) {
    $value = null;
} catch (NoRevisionFoundException $e) {
    $value = null;
}
```

edwines:
Fix #157

### DIFF
--- a/src/SimpleThings/EntityAudit/AuditReader.php
+++ b/src/SimpleThings/EntityAudit/AuditReader.php
@@ -429,6 +429,8 @@ class AuditReader
                                 $value = $this->find($targetClass->name, $pk, $revision, array('threatDeletionsAsExceptions' => true));
                             } catch (DeletedException $e) {
                                 $value = null;
+                            } catch (NoRevisionFoundException $e) {
+                                $value = null;
                             }
 
                             $class->reflFields[$field]->setValue($entity, $value);


bendavies:
@edwines can't you please include a test case for this.

thanks!

bendavies:
@edwines i can't replicate this issue, using a OneToOne like in the issue described.
Can you confirm if this is still any issue?

edwines:
Yes, I pull the most recent version of the EntityAudit and this issue still exists.

Steps to reproduce:

1. Create two Tables like Human and Robot (see #157)
2. Configure the entities in EntityAudit:
```yamel
simple_things_entity_audit:
    audited_entities:
        - AppBundle\Entity\Human
        - AppBundle\Entity\Robot
```
3. Create a Human with a null Robot
4. Call findRevisionHistory and a NoRevisionFoundException will be thrown

nicolasricci:
I can confirm this is still an ongoing issue and this PR fixes this. @bendavies can this be reviewed for merge?

Soullivaneuh:
This way is not right IMO because the associated property will be null like, it does not exists at all.

But this is wrong, the related property exists, just without revision.

We should fallback to the `EntityManager::find` method instead.

See https://github.com/simplethings/EntityAudit/pull/227

DavidBadura:
Closing in favor of #227

Issue #135, opened 2015-06-17
global_ignore_columns config not working


florinutz:
I would like to use this config

```
simple_things_entity_audit:
    global_ignore_columns:
        - created_at
        - updated_at
        - deleted_at
        - created_by
        - updated_by
        - deleted_by
```

(would have been nice to be able to customise ignored columns per entity btw)

But global_ignore_columns doesn't work, all the columns are duplicated.

Is there any way around this?

krossroad:
Having same problem. 'global_ignore_columns' are not working.

OskarStark:
fieldnames are correct

created_at != createdAt

StrikeForceZero:
I was confused by this as well until I logged the value of [$changeset](https://github.com/simplethings/EntityAudit/blob/v0.7/src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php#L111) and saw that my properties were in camelCase but snake_case in the table. Maybe adding a check against a snake_case converted property would be a good idea?

peschee:
A more suitable name for the configuration entry would probably be `global_ignore_fields` since the changeset returned in https://github.com/simplethings/EntityAudit/blob/master/src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php#L191contains the entity field names, not the database table column names.

PR #189, opened 2016-06-27
Rename global_ignore_columns to global_ignore_properties

peschee:
When reading the documentation

```yml
# If you need to exclude some entity properties from triggering a revision use:

simple_things_entity_audit:
    global_ignore_columns:
        - created_at
        - updated_at
```

one would assume that you need to specify the name of your actual database table columns in order to ignore changes for them. However, the `LogRevisionsListener` watches for changes on entity properties (property names) not on the database field names (see #135).

The proposed change is to make the configuration as such:

```yml
simple_things_entity_audit:
    global_ignore_properties:
        - createdAt
        - updatedAt
```

`global_ignore_columns` should rather be called `global_ignore_properties`. I've also updated the tests to reflect this. Since camelCased properties were never tested so the tests never failed.

**Important**: this is a BC-change, so a new version should be tagged accordingly. Or even better: if someone knows a way to support both, but mark the `global_ignore_columns` option as deprecated, feel free to chip in.


tolry:
@peschee I agree, your naming is definitely less error prone.

Keeping this BC compat would be great, has anyone some spare time and the appropriate knowledge to assist @peschee with this?

/cc @smoench @DavidBadura

bendavies:
This seems great, and as previously discussed BC is not a concern at the moment @tolry.
@peschee can you please rebase this?

peschee:
@bendavies I rebased my change on `master`, but I guess now this is wrong? Too many commits it seems :) What's the best way to get rid of those before my change?

bendavies:
@peschee i'm not sure what happened there!

i would have done something like:
```
$ git checkout master
$ git fetch upstream
$ git merge upstream/master
$ git checkout global_ignore_columns_case
$ git rebase master
```

peschee:
@bendavies That's what I did, but without the force push. I guess easiest is to cherry pick my changes into a new branch and make a new PR?

bendavies:
@peschee try the force push! otherwise feel free to make a new PR

peschee:
I have made a new PR #213.

PR #190, opened 2016-06-28
README.md correction

bc-luke:
The audit reader is created from the entity manager.


tolry:
@bc-luke no, that's actually correct, see https://github.com/simplethings/EntityAudit/blob/master/src/SimpleThings/EntityAudit/AuditManager.php#L39

But thanks for your effort still!

PR #192, opened 2016-07-05
AuditedCollection renames indexBy value and then use it. That avoids …

hubipe:
…name collisions between loaded entity identifiers and indexation settings on association.

### DIFF
--- a/src/SimpleThings/EntityAudit/Collection/AuditedCollection.php
+++ b/src/SimpleThings/EntityAudit/Collection/AuditedCollection.php
@@ -448,7 +448,7 @@ class AuditedCollection implements Collection
             $sql .= $this->configuration->getRevisionTypeFieldName().' AS revtype, ';
             $sql .= implode(', ', $this->metadata->getIdentifierColumnNames()).' ';
             if (isset($this->associationDefinition['indexBy'])) {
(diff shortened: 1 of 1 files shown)


tolry:
@hubipe thanks for the contribution!

Did you check, if there is already a test in place to cover this? Also: this would, theoretically, produce a naming collision, if there was a column called __indexBy, correct? Highly unlikely, I know :smile:

bendavies:
@hubipe what exactly does this fix? Can you please submit a testcase to cover this?

hubipe:
This pull request solves problem, when there is collection indexed by one of the primary keys (identifiers). Previous code unset one of the identifiers in keys array, which may problem. With the new implementation, the index by key is renamed to __indexBy and afterwards removed. There might be a name collision only when one of the identifiers is named __indexBy, which as tolry already said, is highly unlikely. Other approach would be to not unset the indexBy key on line 526, if this name collision is detected.
I'm sorry, I can't submit test case, IDK how :(

bendavies:
@hubipe what do your entities look like, which trigger the bug?

DavidBadura:
Can you please create a test so we can reproduce it?

DavidBadura:
If the issue still existis, please create an new PR with a test to reproduce it. Thanks!

PR #195, opened 2016-07-19
Support auditing of entities identified through foreign entities

### LINKED ISSUE #194: Auditing of entities identified through foreign entities does not work.
As of 2.1, Doctrine supports [identity though foreign entities](http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/tutorials/composite-primary-keys.html#identity-through-foreign-entities). EntityAudit cannot currently add revisions for entities whose identifier field is another entity. Consider [this example of a simple derived entity](http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/tutorials/composite-primary-keys.html#use-case-2-simple-derived-identity):

``` php
/**
 * @Entity
 */
class User
{
    /** @Id @Column(type="integer") @GeneratedValue */
    private $id;
}

/**
 * @Entity
 */
class Address
{
    /** @Id @OneToOne(targetEntity="User") */
    private $user;
}
```

User can be audited, but Address cannot because it is identified through User.

jamescdavis:
fix #194

### DIFF
--- a/src/SimpleThings/EntityAudit/AuditReader.php
+++ b/src/SimpleThings/EntityAudit/AuditReader.php
@@ -241,7 +241,7 @@ class AuditReader
             } else if (isset($class->fieldMappings[$idField])) {
                 $columnName = $class->fieldMappings[$idField]['columnName'];
             } elseif (isset($class->associationMappings[$idField])) {
-                $columnName = $class->associationMappings[$idField]['joinColumns'][0];
+                $columnName = $class->associationMappings[$idField]['joinColumns'][0]['name'];
             } else {
                 throw new \RuntimeException('column name not found  for' . $idField);
             }
@@ -362,6 +362,15 @@ class AuditReader
         $keyParts = array();
 
         foreach($class->getIdentifierFieldNames() as $name) {
+            if ($class->hasAssociation($name)) {
+                if ($class->isSingleValuedAssociation($name)) {
+                    $name = $class->getSingleAssociationJoinColumnName($name);
+                } else {
+                    // Doctrine should throw a mapping exception if an identifier
+                    // that is an association is not single valued, but just in case.
+                    throw new \RuntimeException('Multiple valued association identifiers not supported');
+                }
+            }
             $keyParts[] = $data[$name];
         }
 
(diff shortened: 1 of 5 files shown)


DavidBadura:
:+1:

bendavies:
hi @jamescdavis, thanks! can you rebase this please?

jamescdavis:
rebased

bendavies:
Hi @jamescdavis 
Thanks, but there looks to be further issues that need fixing here.
If you add this to the bottom of your test:
```php
        $auditReader = $this->auditManager->createAuditReader($this->em);

        $auditReader->find(get_class($user), $user->getId(), 1);

        $address = $auditReader->find(get_class($address), $address->getUser()->getId(), 2);
        $this->assertEquals($address->getUser(), $user);
```

you'll see it fails, so there are bugs in the audit reader somewhere.

Can you take a look?

jamescdavis:
Hey, @bendavies. That test was failing because my fixtures don't have any getters. :) I was just testing successful persistence, but you're right that I should test the reader too. I'll update the fixtures and add an assert. Also, in the code above, I'm assuming you meant to assign the user found with the audit reader back into $user before the assert?

jamescdavis:
Rebased and added audit reader checks, now failing all builds on travis, but all tests passing for me locally. Going to try to figure out why...

jamescdavis:
Somebody had turned off E_NOTICE on our dev server! I've updated the audit reader and now my tests are passing. Should be good to go now.

DavidBadura:
Sorry for the late response. I'm trying to push the project forward. Can you please rebase your PR? I will merge it immediately if the tests are successful.

jamescdavis:
@DavidBadura looks like you're having build problems with COMPOSER_FLAGS="--prefer-lowest" as well. It seems that composer is taking exceptionally long to calculate the dependencies and Travis is timing out. In my local dev env, composer only takes a few seconds to calculate dependencies, even when using --prefer-lowest. Not sure why it's so slow on Travis.

jamescdavis:
Also, on a related note, even if Travis would let the composer update finish, almost every test fails because composer.json says the minimum required doctrine/orm version is 2.4, and commit 49f9454ca0c49a740a048c4b1fc8a6423cf3f977 makes 2.5 a requirement.

DavidBadura:
Oh, thank you! I have already asked myself why the build fails. I will fix it! Thanks!

PR #197, opened 2016-07-20
Fix auditing of entities with fields that require sql conversion

### LINKED ISSUE #196: Fields that require SQL conversion to convert to database value not converted when saving to the audit table.
When an audited entity contains a field with a custom mapping type that requires calling a SQL function to convert to a value suitable for the database column, EntityAudit is not performing the SQL conversion when saving the field value to the audit table. The original, unconverted value is stored instead.

jamescdavis:
Fix #196


DavidBadura:
:+1:

bendavies:
@jamescdavis can you rebase this please?

jamescdavis:
rebased

bendavies:
thanks @jamescdavis !

PR #198, opened 2016-07-26
fix: use the type of association fields in LogRevisionsListener

mattjmattj:
When building the "UPDATE" query in postFlush, LogRevisionsListener
computes the type of each field to update. When the field was an association
field, the type was not used in the call of Connection::executeQuery, causing
the UPDATE query to fail.

This commit simply adds the computed type in the "types" array


bendavies:
@mattjmattj just here to say thanks for this. I've merged this into my fork for the time being: https://github.com/Owsy/EntityAudit/pull/5

bendavies:
thanks very much @mattjmattj

PR #202, opened 2016-08-16
clean up tests

bendavies:


bendavies:
This package seems pretty unmaintained now. Since I need to make quite a number of changes, i'm going to maintain my own fork for now.

PR #205, opened 2016-08-17
Travis improvements

bendavies:
some improvements to the travis config.

/cc @tolry

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ language: php
 
 sudo: false
 
+branches:
+  only:
+    - master
+
 php:
   - 5.3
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,22 +2,34 @@ language: php
 
 sudo: false
 
+branches:
+  only:
+    - master
```
DavidBadura: why do you want only test the master branch?

bendavies: it stops PR's building twice if the PR source is in this repository.

DavidBadura: Ah, okay :-)

PR #210, opened 2016-08-19
Provide a way to customize the revision username.

### LINKED ISSUE #187: User identifier may change multiple times during execution - rely on a dynamic factory for it instead?
During an app's lifecycle, many things can go wrong with auth + business logic.

 * email auto-login + action (via JWT token or similar) causes anonymous -> logged in upgrade + business interaction
 * `sudo` -alike actions cause user privilege change mid-request
 * auto-logout caused by business logic (violations, usually), lead to logged in -> anonymous downgrade

Therefore, `AuditConfiguration#setCurrentUsername()` seems to be insufficient to deal with that.

Assuming (big assumption tho!) that there is somewhere a storage keeping the current logged in user in memory, it is easy to change the logic to accomodate this, by having a `AuditConfiguration#setCurrentUsernameProvider(callable $usernameProvider)`. That way, each time the username is required, the provider is called, making sure the interactions are executed with the correct role at every transaction within the lifecycle of a request.

bendavies:
This fixes #187

This provide a more flexible way to provide the username for the current revision.

The default way comes from the security context token (as previously) but is evaluated lazily, not once at the beginning of the request as was done previously, which #187 highlights.

The username resolving can also be customized by configuring a custom service.

cc @Ocramius @simplethings/entityaudit


src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
```diff
@@ -309,6 +309,16 @@ private function getRevisionId()
     }
 
     /**
+     * @return string|null
+     */
+    private function getUsername()
+    {
+        $callable = $this->config->getUsernameCallable();
```
smoench: @bendavies `$callable` could be `null`

src/SimpleThings/EntityAudit/AuditConfiguration.php
```diff
@@ -27,15 +27,49 @@
...
+    public function __construct(
+        array $auditedEntityClasses = array(),
+        array $globalIgnoreColumns = array(),
+        $tablePrefix = '',
+        $tableSuffix = '_audit',
+        $revisionTableName = 'revisions',
+        $revisionIdFieldType = 'integer',
+        $revisionFieldName = 'rev',
+        $revisionTypeFieldName = 'revtype',
+        callable $usernameCallable = null
```
smoench: @bendavies "Callbacks can be denoted by callable type hint as of PHP 5.4."  -> http://php.net/manual/en/language.types.callable.php

src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
bendavies: thanks. fixed.

src/SimpleThings/EntityAudit/AuditConfiguration.php
bendavies: hah, damn.

bendavies: thanks, fixed.

src/SimpleThings/EntityAudit/DependencyInjection/Configuration.php
```diff
@@ -24,12 +24,10 @@ public function getConfigTreeBuilder()
                 ->scalarNode('revision_type_field_name')->defaultValue('revtype')->end()
                 ->scalarNode('revision_table_name')->defaultValue('revisions')->end()
                 ->scalarNode('revision_id_field_type')->defaultValue('integer')->end()
-
-                ->arrayNode('listener')
+                ->arrayNode('service')
                     ->addDefaultsIfNotSet()
-                        ->children()
-                            ->booleanNode('current_username')->defaultTrue()->end()
```
OskarStark: this is a bc break and should be deprecated first

src/SimpleThings/EntityAudit/EventListener/CreateSchemaListener.php
```diff
@@ -109,7 +109,7 @@ public function postGenerateSchema(GenerateSchemaEventArgs $eventArgs)
             'autoincrement' => true,
         ));
         $revisionsTable->addColumn('timestamp', 'datetime');
-        $revisionsTable->addColumn('username', 'string');
+        $revisionsTable->addColumn('username', 'string')->setNotnull(false);
```
OskarStark: this is a BC break which should be deprecated

src/SimpleThings/EntityAudit/Request/CurrentUserListener.php
```diff
@@ -1,70 +0,0 @@
...
-use Symfony\Component\HttpKernel\HttpKernelInterface;
-use Symfony\Component\HttpKernel\Event\GetResponseEvent;
-use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
-use Symfony\Component\Security\Core\SecurityContext;
-use SimpleThings\EntityAudit\AuditConfiguration;
-
-/**
- * Inject the SecurityContext username into the AuditConfiguration as current username.
- */
-class CurrentUserListener
```
OskarStark: removing a class is BC break

bendavies:
@OskarStark thanks for your comments, but this bundle is still pre release so there is no BC to uphold: http://semver.org/#spec-item-4

OskarStark:
> @OskarStark thanks for your comments, but this bundle is still pre release so there is no BC to uphold: http://semver.org/#spec-item-4

you are right in general, but this bundle is used by many of them and introducing BC in master is fine, but if you don't deprecate them its hard to upgrade from 0.9.x to 1.x.x and that is one reason why people DON'T do the upgrade. the have no path to go and ensure everything is working as expected...

IMO if pre release or not this is a very hard change which is not documented in an UPGRADE file which i could follow at least (if you don't do the deprecation)

btw the PR itself is nice ;-)

bendavies:
@OskarStark thanks. the CHANGELOG will certainly be updated and I agree with you that an UPGRADE file could be created so the breaking changes can be documented and the upgrade path made easier.

However, in this instance I don't agree with you that the changes are so severe that it would prevent users from upgrading. They are pretty minor and would require minimal for the implementor.

@tolry @DavidBadura some input from you both would be appreciated here.

src/SimpleThings/EntityAudit/AuditConfiguration.php
```diff
@@ -27,15 +27,49 @@
...
+     * @param array $globalIgnoreColumns
+     * @param $tablePrefix
+     * @param $tableSuffix
+     * @param $revisionTableName
+     * @param $revisionFieldName
+     * @param $revisionTypeFieldName
+     * @param $revisionIdFieldType
+     * @param callable $usernameCallable
+     */
+    public function __construct(
```
Ocramius: Instead of using `__construct` here, I suggest moving the defaults back into the properties, and then defining a named constructor. The reason for doing so is that additional fields will always come up, and sometimes you may want to make them required. In order to do so, you do not want to change `__construct` (which would result in a BC break), but deprecate the old named constructor and add a new one.

src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
```diff
@@ -309,6 +309,16 @@ private function getRevisionId()
...
+
+        return $callable ? $callable() : null;
```
Ocramius: What if the "old style" username is required (where username was just assigned?) 

Suggestion: make the `$callable` return the configured username by default.

src/SimpleThings/EntityAudit/AuditConfiguration.php
```diff
@@ -124,14 +158,25 @@ public function createMetadataFactory()
...
-        $this->currentUsername = $username;
+        // php 5.3 compat
+        if (null !== $usernameCallable && !is_callable($usernameCallable)) {
+            throw new \InvalidArgumentException(sprintf(
+                'Username Callable must be callable. Got: %s',
+                is_object($usernameCallable) ? get_class($usernameCallable) : gettype($usernameCallable)
+            ));
+        }
+
+        $this->usernameCallable = $usernameCallable;
```
Ocramius: Suggestion: `$usernameCallable ?: function () { return $this->currentUsername; }`

src/SimpleThings/EntityAudit/AuditConfiguration.php
```diff
@@ -124,14 +158,25 @@ public function createMetadataFactory()
         return new Metadata\MetadataFactory($this->auditedEntityClasses);
     }
 
-    public function setCurrentUsername($username)
```
Ocramius: Removing `setCurrentUsername` is a BC break. It is simple to work around this by making this internally do:

```php
/**
 * @deprecated
 */
public function setCurrentUsername($username)
{
    $this->setUsernameCallable(function () use ($username) {
        return $username;
    });
}
```

src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
Ocramius: Moved the suggestion to the `Configuration` comments

tests/SimpleThings/Tests/EntityAudit/BaseTest.php
```diff
@@ -120,10 +120,10 @@ protected function getEntityManager()
      */
     protected function getAuditManager()
     {
-        $auditConfig = new AuditConfiguration();
-        $auditConfig->setCurrentUsername('beberlei');
```
Ocramius: Changes can be reverted if suggestions above are applied

tests/SimpleThings/Tests/EntityAudit/BaseTest.php
```diff
@@ -120,10 +120,10 @@ protected function getEntityManager()
...
-        $auditConfig->setAuditedEntityClasses($this->auditedEntities);
-        $auditConfig->setGlobalIgnoreColumns(array('ignoreme'));
+        $auditConfig = new AuditConfiguration($this->auditedEntities, array('ignoreme'));
+        $auditConfig->setUsernameCallable(function () {
+            return 'beberlei';
```
Ocramius: A new test that sets `uniqid` as a callback would be best here (showing that the username changes over multiple flush calls)

src/SimpleThings/EntityAudit/AuditConfiguration.php
bendavies: good idea!

tests/SimpleThings/Tests/EntityAudit/BaseTest.php
bendavies: keeping this as `setCurrentUsername` is deprecate as you suggest above.

src/SimpleThings/EntityAudit/AuditConfiguration.php
bendavies: yep, great.

tests/SimpleThings/Tests/EntityAudit/BaseTest.php
bendavies: ...and done!

bendavies:
thanks @Ocramius - addressed!

src/SimpleThings/EntityAudit/AuditConfiguration.php
bendavies: kept and deprecated setCurrentUsername instead

bendavies: kept and deprecated setCurrentUsername instead

Ocramius:
@bendavies awesome, thanks!

OskarStark:
> especially with an UPGRADE file.
in this case yes ;-)

src/SimpleThings/EntityAudit/EventListener/CreateSchemaListener.php
DavidBadura: why it is a BC break? Allow NULL isn't a BC break and your application will not break down.

bendavies:
@DavidBadura @tolry I'm going to merge this on monday if there are no more comments/objections,

src/SimpleThings/EntityAudit/EventListener/CreateSchemaListener.php
OskarStark: oh in this case you are right, but changing a schema could be a problem (not in this case) if the schema change is not recognized and handled on production

DavidBadura:
:+1: for this. but we should tag a new version before we merge this :-)

bendavies:
@DavidBadura will do.

DavidBadura:
@bendavies thank you! :)

bendavies:
@DavidBadura thankyou!

PR #212, opened 2016-08-22
Drop php < 5.6

### LINKED ISSUE #207: RFC: drop support for PHP < 5.6
Hey,

It looks like this library still supports all php versions back to 5.3.
Since versions < 5.6 are end of life, i propose dropping support for them.

cc @simplethings/entityaudit

bendavies:
Drop support for EOL php versions: 5.3/5.4/5.5
resolves #207 

cc @DavidBadura @tolry

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,6 @@ cache:
     - $HOME/.composer/cache
 
 php:
-  - 5.3
-  - 5.4
-  - 5.5
   - 5.6
   - 7.0
   - nightly
@@ -38,7 +35,7 @@ script: ./vendor/bin/phpunit -c tests/travis/$DB.travis.xml
 matrix:
   fast_finish: true
   include:
-    - php: 5.3
+    - php: 5.6
       env: COMPOSER_FLAGS="--prefer-lowest" DB=sqlite
     - php: 7.0
       env: NAME="MySQL 5.6" DB=mysql
--- a/README.md
+++ b/README.md
@@ -96,11 +96,13 @@ use Doctrine\ORM\EntityManager;
 use Doctrine\Common\EventManager;
 use SimpleThings\EntityAudit\AuditConfiguration;
 use SimpleThings\EntityAudit\AuditManager;
+use SimpleThings\EntityAudit\Tests\ArticleAudit;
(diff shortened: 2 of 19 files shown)


tolry:
As discussed before, I would suggest tagging the current version with 1.0 and moving on from there with 2.* with larger changes, like these. Sounds good or alternative suggestions?

bendavies:
@tolry i didn't realise you want to release v1 first. you only mentioned tagging previously.
I'd rather hold off v1.0 for now as there are still a few confirmed bugs and good PRs opening that i'd like to get merged first.

We can hold off on this PR for now if you prefer.

tolry:
We could make due without the 1.0 tagging / 2.* jump. But I myself do not like 0.* releases very much, not a lot of rules defined in SemVer for those.

If we end up having larger changes / BC breaks in 0.* releases, it would be really important to update the changelog precisely (as seems to be done currently). I would think, that tagging the current version with 1.0 and releasing 2.* directly after that, might make things clearer for most users. But I am open to different approaches.

Thanks a lot for your hard work so far, by the way!

bendavies:
@tolry I do agree with all these `0.*` releases. Ultimately I'm happy to whatever you ,@smoench and @DanielBadura  want to do though, as you've been working on this longer than me.

DavidBadura:
this is a very difficult question. we can release this state as 1.0 and merge all bc-breaks into master to prepare 2.0. but then we need support two versions: 1.0 and master (2.0). 

some bugfixes introduce bc breaks, so in which one should we merge this? should we stabilize this state before we beginn to working on 2.0?

the jump from 0.x to 1.x is also a major release, so bc breaks should be allowed. my opinion is to release 1.0 after this and all other bc breaks to have a clean 1.0. but i'm also fine with releasing 1.0 before merging this.

rvanlaak:
Does the upcoming release also bump other requirements? Like the symfony framework? If so, then there already are breaks and I'd propose to also break the rest ;-) 

As support on this bundle is not really that active (no problem!), supporting two versions will not happen right? :+1:

DavidBadura:
Closing in favor of #239

PR #213, opened 2016-08-23
Rename global_ignore_columns to global_ignore_properties

peschee:
**Replaces #189 because of a bad rebase.**

When reading the documentation

``` yml
# If you need to exclude some entity properties from triggering a revision use:

simple_things_entity_audit:
    global_ignore_columns:
        - created_at
        - updated_at
```

one would assume that you need to specify the name of your actual database table columns in order to ignore changes for them. However, the `LogRevisionsListener` watches for changes on entity properties (property names) not on the database field names (see #135).

The proposed change is to make the configuration as such:

``` yml
simple_things_entity_audit:
    global_ignore_properties:
        - createdAt
        - updatedAt
```

`global_ignore_columns` should rather be called `global_ignore_properties`. I've also updated the tests to reflect this. Since camelCased properties were never tested so the tests never failed.

**Important**: this is a BC-change, so a new version should be tagged accordingly. Or even better: if someone knows a way to support both, but mark the `global_ignore_columns` option as deprecated, feel free to chip in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,9 +70,9 @@ If you need to exclude some entity properties from triggering a revision use:
 #####app/config/config.yml
 ```yml
 simple_things_entity_audit:
-    global_ignore_columns:
-        - created_at
-        - updated_at
+    global_ignore_properties:
+        - createdAt
+        - updatedAt
 ```
 
 ###Creating new tables
@@ -103,9 +103,9 @@ $auditconfig->setAuditedEntityClasses(array(
     'SimpleThings\EntityAudit\Tests\UserAudit'
 ));
 
-$auditconfig->setGlobalIgnoreColumns(array(
-    'created_at',
-    'updated_at'
+$auditconfig->setGlobalIgnoreProperties(array(
+    'createdAt',
+    'updatedAt'
 ));
 
 $evm = new EventManager();
--- a/src/SimpleThings/EntityAudit/AuditConfiguration.php
+++ b/src/SimpleThings/EntityAudit/AuditConfiguration.php
@@ -28,7 +28,7 @@ use Doctrine\ORM\Mapping\ClassMetadataInfo;
 class AuditConfiguration
 {
     private $auditedEntityClasses = array();
-    private $globalIgnoreColumns = array();
+    private $globalIgnoreProperties = array();
     private $tablePrefix = '';
     private $tableSuffix = '_audit';
     private $revisionTableName = 'revisions';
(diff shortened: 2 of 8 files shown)


bendavies:
@smoench @DavidBadura @tolry this seems sensible.

tolry:
As mentioned before, :+1: generally.

But this will break configuration for lots of users, therefore I would prefer some sort of BC layer to still accept the old parameters. Anyone know an elegant solution for that? config key aliases or something like that?

bendavies:
yep agree, will try find a nice way...

bendavies:
@peschee can you try putting `global_ignore_columns` back in the config, add an `->info(...)` node to mark it as deprecated and replaced by `global_ignore_properties`. Set the parameter from which ever is configured.

Then validate that they are both not set at the end?  (not tested)
```
 ->validate()
    ->ifTrue(function ($config) {
        return config['global_ignore_columns'] && $config['global_ignore_properties'];
    })
    ->thenInvalid('The `global_ignore_columns` and `global_ignore_properties` options are mutually exclusive. Please use `global_ignore_properties`.')
>end();
```

bendavies:
With something like this?

```
->beforeNormalization()
    ->ifTrue(function ($v) {return isset($v['global_ignore_columns']);})
    ->then(function ($v) {
        $v['global_ignore_properties'] = $v['global_ignore_columns'];
        unset($v['global_ignore_columns']);

        return $v;
    })
->end()
```

DavidBadura:
:+1: looks good

peschee:
@bendavies Does this look good?

https://github.com/peschee/EntityAudit/commit/31b0c98a01cc0f9225bbe1c7fc1376e0b03f9720

I have no experience with writing configuration values code. Would you care to take a look? We should also add a test for the validation you proposed (in case both `global_ignore_columns` and `_properties` are present.

Soullivaneuh:
> But this will break configuration for lots of users

0.x versions can be BC break, an upgrade note would be enough IMHO.

But if the BC way it's already here, then :+1:.

tolry:
@Soullivaneuh technically I agree, SemVer allows all kinds of breaks in 0.x

Difference here is, that 0.x is in production use in lots and lots of projects (EntityAudit is the Recommended versioning method for SonataAdminBundle). Therefore this is de facto not a real 0.x version, but more of a 1.x

DavidBadura:
thank you!

PR #222, opened 2016-09-14
Support for entities annotations

### LINKED ISSUE #20: [Request] Annotation support and lasted revision
Is possible add annotation support and a method to get the lasted revision?

tseho:
I added an Auditable annotation for entities.

```
<?php
...
use SimpleThings\EntityAudit\Mapping\Annotation as Audit;

/**
 * @ORM\Entity()
 * @Audit\Auditable
 */
class TestAudit {
```

Annotated entities are merged into the audited_entities array.

IMO, we should advise to use annotations instead of the global configuration in app/config/config.yml, this is why I added a deprecated message when the configuration key "audited_entities" is set.

For avoiding BC breaks, I had to wrap the AuditConfiguration service with a factory, I updated the SimpleThingsEntityAuditExtensionTest correspondingly.

---

Tests are passing, if this is something you would want to merge, I can add more tests on annotated entities.

---

Fixes #20

### DIFF
--- a/src/SimpleThings/EntityAudit/AuditConfigurationFactory.php
+++ b/src/SimpleThings/EntityAudit/AuditConfigurationFactory.php
@@ -3,13 +3,20 @@
 namespace SimpleThings\EntityAudit;
 
 use Doctrine\Common\Annotations\Reader;
-use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ORM\Configuration;
 
 class AuditConfigurationFactory
 {
     const ANNOTATION_AUDITABLE = 'SimpleThings\\EntityAudit\\Mapping\\Annotation\\Auditable';
 
+    /**
+     * This is the factory method used by the Symfony DependencyInjection
+     *
+     * @param Configuration $doctrineConfiguration
+     * @param Reader $reader
+     * @param array $auditedEntities
+     * @return AuditConfiguration
+     */
     public function createAuditConfiguration(Configuration $doctrineConfiguration, Reader $reader, $auditedEntities = array())
     {
         $entities = $doctrineConfiguration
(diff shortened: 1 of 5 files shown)


rvanlaak:
Having multiple configuration formats in my opinion always is great. It allows users that for instance define their entities via XML to choose their appropriate way of configuring the bundle.

In other words, although I prefer annotations myself, I don't think that the current configuration format should be deprecated.

jamescdavis:
I think this is a great idea and would probably use annotations, however, I do agree with @rvanlaak that the current configuration format should not be deprecated.

tseho:
No problem, this was just a warning in the configuration, I will revert it.

rvanlaak:
Although I think there should be added a check to not mix-up the two configuration formats...

tseho:
A warning if there is entities registered by both the config and the
annotations ?

(Btw, it's something I tried and it's working, you can use both)

rvanlaak:
Oh if it is working then I'm not sure, mixing up config formats sounds scary to me ;) Maybe @DavidBadura  can elaborate on it.

tseho:
I agree that using the config and the annotations at the same time seems
dirty. It's why I added the deprecated warning but we can do something else
instead.

I wanted to promote annotations because it will allow us to have new
features (like the one discussed in #223)

s001dxp:
Being able to configure based on ones needs is important. Nearly every bundle out there gives one the opportunity to choose.

tseho:
@DavidBadura sure. I will rebase and remove the deprecated warning during the weekend. I don't have time for adding additionnal tests right now.

tseho:
@DavidBadura done

DavidBadura:
Thank you, I will continue to work on it.

PR #229, opened 2016-10-08
Update LogRevisionsListener.php

Petah:
Fix duplicate key error when deleting a mix of proxies and non proxies.

### DIFF
--- a/src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
+++ b/src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
@@ -469,7 +469,7 @@ class LogRevisionsListener implements EventSubscriber
         return implode(
             ' ',
             array_merge(
-                array(get_class($entity)),
+                array(\Doctrine\Common\Util\ClassUtils::getRealClass(get_class($entity))),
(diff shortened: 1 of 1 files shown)


tolry:
@Petah looks good to me - could you add a test to specifically reproduce this behaviour?

src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
```diff
@@ -469,7 +469,7 @@ private function getHash($entity)
         return implode(
             ' ',
             array_merge(
-                array(get_class($entity)),
+                array(\Doctrine\Common\Util\ClassUtils::getRealClass(get_class($entity))),
```
Soullivaneuh: This can be simplified by using `ClassUtils::getClass($entity)` instead.

Ref: https://github.com/doctrine/common/blob/v2.7.1/lib/Doctrine/Common/Util/ClassUtils.php#L56-L59

Also, you may import the namespace to get a cleaner code.

DavidBadura: done 0ad8a114372388a509922bca432b414b20f7deff

Issue #182, opened 2016-04-13
Any reason for private vs protected properties?


jtreminio:
Setting everything to private makes it extremely difficult to extend.

Your thoughts on a PR to convert `private` to `protected`?

tolry:
> Setting everything to private makes it extremely difficult to extend.

Generally speaking: that's the point :wink: 

When developing a library, making something private, is the only way in php to declare those parts as "not meant to be extended". Everything that can be extended is a potential BC break for users, if the library gets refactored. You might take a look at the doctrine core library, most of the classes and properties are private there too, for the same reason.

Extendability (is that a word? I hope so) should be explicite, otherwise developers forget about it at some point and break code they do not know about.

Since I am not very familiar with the EntityAudit code in detail: do you have examples, where you would like to be able to extend EntityAudit?

There was a discussion in #83 about enabling developers to hook into the system by the means of events, which I personally think is a great way of extending. Having real life examples would probably help that discussion.

jtreminio:
I'm familiar with the train of thought of making everything private by default. However, in the better examples of this, those projects provide hooks as you mentioned. This one has neither.

Our use case can't be unique, but:

* I'm working at a healthcare company that requires strict control over private identifiable information (PII). Having `revisions.username` use the user's username (or email address) would be a big red flag. We need to either encrypt it or better yet use user ID because a user's email address or username may very well change in the future through profile edits, but ID should always be immutable.
* We want to keep track of the order records are added/changed/deleted in a single transaction, so we've added a `rev_order` column to each `*_audit` table.
* We need to add a `revisions.change_reason` column for record keeping.

As far as I can tell, none of these things are possible without forking the project and overriding much of the code.

cordoval:
@jtreminio did you have your mods somewhere? thanks in advance

cordoval:
nvm https://github.com/jtreminio/EntityAudit/commits/forked

DavidBadura:
Like @tolry says: We prefer to provide hooks/events. I try to integrate it with version 2.0 (#83)

PR #233, opened 2016-10-22
Enable configurable column value replace

oconnedk:
It might be a limited use-case, but there are times that you do not want to store a revision history for a certain sensitive field (or fields).
Password, for example, if you were crazy enough to store them in plain - or maybe you are encrypting them, but they are not salted etc.
This PR includes unit tests for the feature and documentation in the  [README.md](https://github.com/oconnedk/EntityAudit/blob/enable_column_value_replace/README.md#appconfigconfigyml-2) file.


DavidBadura:
thanks for your work, but we want to offer hooks so you can customize the data. See #83 and #182. I try to integrate it with version 2.0

PR #239, opened 2017-01-05
Drop php < 5.6 & symfony < 2.8

### LINKED ISSUE #207: RFC: drop support for PHP < 5.6

DavidBadura:
replace #212

We shouldn't change the array syntax etc. everywhere so that we don't have too much conflicts.

thanks @bendavies

closed #207

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,11 +11,9 @@ cache:
     - $HOME/.composer/cache
 
 php:
-  - 5.3
-  - 5.4
-  - 5.5
   - 5.6
   - 7.0
+  - 7.1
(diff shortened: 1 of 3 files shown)


rvanlaak:
As of this merge we _can_ change the array syntax right?

DavidBadura:
yes, you *can* use the short array syntax for new code. But we should keep the long array syntax in old code until we merge the most pr.

PR #243, opened 2017-01-08
[Refactoring] Use dbal query builder

smoench:
| Q             | A
| ------------- | ---
| Bug fix?      | probably
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Has tests?    | yes
| Fixed tickets | -

Now it would be easier to identify and reduce code duplication.


DavidBadura:
:+1: LGTM

Do we have any performance problems? I don't think so.

DavidBadura:
thank you!

Issue #230, opened 2016-10-11
SQL Server and sequence name


dlegatt:
I have to use SQL Server for my database and am running into an error.  In LogRegisionsListener::getRevisionId(), there is a check to see if the platform supports sequences, which the SqlServer2012 platform does, but then it calls getIdentitySequenceName, which the SqlServerPlatform does not implement, so an exception is thrown.

DavidBadura:
I think it is an issue in doctrine dbal. Or can you tell us how to solve it?

dlegatt:
It probably is a doctrine dbal issue.  I worked around it in my fork by removing the sequenceName generation starting on line 309 and instead have
`$this->revisionId = $this->conn->lastInsertId();`

DavidBadura:
Can you create a PR so we can run our tests?

dlegatt:
I have very little experience with GitHub outside of hosting my own projects.  Can you tell me what I need to do to make a PR for you?

dlegatt:
I made a PR, but it looks like it makes the tests fail

smoench:
This is an issue with doctrine dbal and is maybe related to https://github.com/doctrine/dbal/issues/2467.
It should be fixed in dbal not in entity-audit.

DavidBadura:
it should be solved in dbal (doctrine/dbal#2467)

PR #245, opened 2017-01-10
Updated LogRevisionsListener.php

dlegatt:
Removed sequence name generation to fix error with SqlServer platform

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Has tests?    | no
| Fixed tickets | #230

### DIFF
--- a/src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
+++ b/src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
@@ -331,12 +331,8 @@ class LogRevisionsListener implements EventSubscriber
                     Type::STRING,
                 )
             );
-
-            $sequenceName = $this->platform->supportsSequences()
-                ? $this->platform->getIdentitySequenceName($this->config->getRevisionTableName(), 'id')
-                : null;
-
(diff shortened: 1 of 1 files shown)


DavidBadura:
Thanks! I am wondering why only one environment (postgres + 5.6) fails :-/ I have restarted the tests...

DavidBadura:
I think the low dependencies are the cause

DavidBadura:
Can you increase the dbal requirements on patch level until the tests are green?

from `~2.5` to:

* `^2.5.1`
* `^2.5.2`
* ...

I think dbal has solved it, but I don't know in which version.

--------

You need only change the composer.json, commit and push it. Everything else happens automatically.

smoench:
👎 #230 is an issue with dbal and sql server.
I think pgsql (in older versions) is requiring a sequence name for last inserted id.

IMHO: if a platform is supporting sequence names they should be used.

DavidBadura:
Ah, now I understand. Sorry @dlegatt but it's an error in dbal (https://github.com/doctrine/dbal/issues/2467). We should keep sequence support for other platforms.

PR #247, opened 2017-01-11
Check getGlobalIgnoreProperties before add an entity to the extraUpda…

marphi:
…tes array

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Has tests?    | no
| Fixed tickets | #... 

GlobalIngoreProperties should be checked before added an entity to `extraUpdates` during the onFlush event to prevent create unnecessary revisions.

### DIFF
--- a/src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
+++ b/src/SimpleThings/EntityAudit/EventListener/LogRevisionsListener.php
@@ -294,6 +294,19 @@ class LogRevisionsListener implements EventSubscriber
                 continue;
             }
 
+            // get changes => should be already computed here (is a listener)
+            $changeset = $this->uow->getEntityChangeSet($entity);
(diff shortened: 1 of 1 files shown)


DavidBadura:
Thanks! Can you write a test for this?

marphi:
I've started write test but I've found the base problem in my case.
This PR prevents to execute unnecessary SQL, when all properties have been ignored.
SQL Like:
'UPDATE ArticleAudit_audit SET ignoreMe = ? WHERE rev = ? AND id = ?' with non exists revision ID.

I've seen this sql in my log and I thought that new revision was created but it was only unnecessary SQL with non exists revision Id.

It will be very hard to write test for this case.

DavidBadura:
Ah, in this case, it's okay! Thanks!

PR #249, opened 2017-01-11
Manage associations on AuditReader::diff

### LINKED ISSUE #123: Foreign key diff
Hi everyone,

Is there any change to see also foreign key changes?

Thank you in advance!

Soullivaneuh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Has tests?    | WIP

Fixes #123 

- [x] Manage single relations (ManyToOne / OneToOne)
- [x] Manage collections (OneToMany / ManyToMany) -> Must be *ignored*
- [x] Add some tests
- [x] Remove not used `ArrayDiff` class

I have to move the `ArrayDiff` class logic onto the AuditReader, because we need the metadata factory.

Plus, this class was used only on this part of the code. I think this is not necessary to separate this logic and we can remove safely this class.

Returned values on the diff are the object, but comparisons are done with identifiers only, because we check for the relation diff, not the content of the related object.

@DavidBadura How should we manage collections? Compare length and identifier inside? Just ignore them?


DavidBadura:
For collections we can return an array with added and removed elements?

```json
{
  "added": [5, 6],
  "removed": [1]
}
```

Better we should create a `DiffCollection` object, with added and removed elements.

Soullivaneuh:
> For collections we can return an array with added and removed elements?

Well it will produce a completely different API than `old`, `new` and `same.

The other solution would be to keep the `old/new/same` logic but with old and new (or same) collections on it.

> Better we should create a `DiffCollection` object, with added and removed elements.

Same things. Looks a good idea but in this case we should do the same for the `old/new/same` system with, why not, a `DiffInterface`. But I think this is an idea for an another MR. :wink:

Soullivaneuh:
@DavidBadura I think collections must be simply removed from the diff.

In any version of the entity, because collections are from another table, the collection size and content will always be the same.

So the diff is not possible.

DavidBadura:
okay, do it :)

Soullivaneuh:
@DavidBadura all tasks done. Ready for review. :+1:

DavidBadura:
LGTM, thanks!

Soullivaneuh:
@DavidBadura Travis is good!

Do you have any ETA for the 2.0 release?

PR #250, opened 2017-01-13
Fix OneToOne reverse relation Doctrine fallback query

Soullivaneuh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Has tests?    | yes

The foreign column name was used on a classic Doctrine find.

Two issues:

* This method expects an identifier. So we have to get the repository to use findOneBy method instead.
* The findOneBy method expects fieldName, not column name. So primary field array is introduced.

I target the 1.0 branch because it's a bugfix of introduced feature in #227.

You can push the new patch release after this PR. :+1:

### DIFF
--- a/src/SimpleThings/EntityAudit/AuditReader.php
+++ b/src/SimpleThings/EntityAudit/AuditReader.php
@@ -420,18 +420,21 @@ class AuditReader
                 //print_r($targetClass->discriminatorMap);
                 if ($this->metadataFactory->isAudited($assoc['targetEntity'])) {
                     if ($this->loadAuditedEntities) {
+                        // Primary Key. Used for audit tables queries.
                         $pk = array();
+                        // Primary Field. Used when fallback to Doctrine finder.
+                        $pf = array();
 
                         if ($assoc['isOwningSide']) {
                             foreach ($assoc['targetToSourceKeyColumns'] as $foreign => $local) {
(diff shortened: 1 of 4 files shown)


Soullivaneuh:
BTW, this PR must be merged to master after that.

DavidBadura:
thanks!

DavidBadura:
https://github.com/simplethings/EntityAudit/releases/tag/v1.0.1

Soullivaneuh:
Thank you @DavidBadura !

Soullivaneuh:
@DavidBadura I think the tag was on the wrong place, including BC breaks.

For `global_ignore_columns` configuration:

* 1.0.0: https://github.com/simplethings/EntityAuditBundle/blob/v1.0.0/src/SimpleThings/EntityAudit/DependencyInjection/Configuration.php#L18
* 1.0.1: https://github.com/simplethings/EntityAuditBundle/blob/v1.0.1/src/SimpleThings/EntityAudit/DependencyInjection/Configuration.php#L18

This modification should be on v2.0, right?

DavidBadura:
@Soullivaneuh Thanks for the hint! Version 1.0.2 should solved this.

Soullivaneuh:
Thanks @DavidBadura. I just upgraded to 1.0.2, works correctly. :+1:

Issue #223, opened 2016-09-16
Ignore fields by entity


Soullivaneuh:
Currently, we can setup global ignored columns.

It would be great to be able to specify ignored columns by entity.

Or/And maybe the reverse: Specify which fields we want to track.

tseho:
I have been looking into it and it's something I want to include with the next milestone (1.0 ?) cc @tolry 

I had in mind to use the same approach as the jms serializer: [exclusion strategies](http://jmsyst.com/libs/serializer/master/cookbook/exclusion_strategies#general-exclusion-strategies)

Soullivaneuh:
Looks like a good idea. :+1:

Soullivaneuh:
@tseho Waiting that, it would be quite hard to override the listener:

```php
    public function postUpdate(LifecycleEventArgs $eventArgs)
    {
        // onFlush was executed before, everything already initialized
        $entity = $eventArgs->getEntity();

        $class = $this->em->getClassMetadata(get_class($entity));
        if (! $this->metadataFactory->isAudited($class->name)) {
            return;
        }

        // get changes => should be already computed here (is a listener)
        $changeset = $this->uow->getEntityChangeSet($entity);
        foreach ($this->config->getGlobalIgnoreColumns() as $column) {
            if (isset($changeset[$column])) {
                unset($changeset[$column]);
            }
        }

        // if we have no changes left => don't create revision log
        if (count($changeset) == 0) {
            return;
        }

        $entityData = array_merge($this->getOriginalEntityData($entity), $this->uow->getEntityIdentifier($entity));
        $this->saveRevisionEntityData($class, $entityData, 'UPD');
    }
```

In this case, I just want to manage the changeset, but I have to override and copy/paste the whole method.

Maybe should we provide a `getChangeSet` method for now?

Soullivaneuh:
BTW, we can't override, needed properties are privates...

Soullivaneuh:
I had to copy the private properties for now. Here is a draft of what I'm doing:

```php
namespace AppBundle\Listener;

use AppBundle\Entity\Ticket;
use Doctrine\ORM\Event\LifecycleEventArgs;
use Doctrine\ORM\Event\OnFlushEventArgs;
use SimpleThings\EntityAudit\AuditManager;
use SimpleThings\EntityAudit\EventListener\LogRevisionsListener as BaseLogRevisionsListener;

/**
 * @author Sullivan Senechal <soullivaneuh@gmail.com>
[...]

Issue #140, opened 2015-07-08
dbal connection hardcoded to default


malotor:
I have a problem with database conections. I have a sqlite database for development and a sqlserver database for other enviroments.

I have this config file

config.yml

```
doctrine:
    dbal:
        default_connection:     "%database_default_connection%"
        connections:
            default:
                driver:         "%database_driver%"
                path:           "%database_path%"
            sqlserver:
                driver_class:   "%database_class%"
                host:           "%database_host%"
                port:           "%database_port%"
                dbname:         "%database_name%"
                user:           "%database_user%"
                password:       "%database_password%"
                charset:        UTF8

```

parameters.yml.dev

```
parameters:
    database_default_connection : default
```

parameters.yml.prod

```
parameters:
    database_default_connection : sqlserver
```

In production enviroment the audit bundle does`t work. I think this is due to de bundle always use de hardcode "default" connection.

In fact, if i rename the "default" connection name to "sqlite" and change

```
parameters:
    database_default_connection : sqlite
```

It have this error during "composer install"

he Doctrine connection "default" referenced in service "simplethings_entityaudit.log_revisions_l│
istener" does not exist. Available connections names: sqlite, sqlserve

I see the service and in fact, it have hardcode the connection default

```
<service id="simplethings_entityaudit.log_revisions_listener" class="SimpleThings\EntityAudit\EventListener\LogRevisionsListener">
            <argument id="simplethings_entityaudit.manager" type="service" />
            <tag name="doctrine.event_subscriber" connection="default" />
        </service>
```

Any suggestion?

v-bartusevicius:
Same problem here.
I think best solution would be to remove `connection="default"` as this way doctrine will use the default connection from config automatically.

v-bartusevicius:
As working workaround I could suggest to create a compiler pass like this:
```
class OverrideEntityAuditBundleConnectionCompilerPass implements CompilerPassInterface
{
    private $servicesToOverride = array(
        'simplethings_entityaudit.log_revisions_listener',
        'simplethings_entityaudit.create_schema_listener',
    );

    public function process(ContainerBuilder $container)
    {
        foreach ($this->servicesToOverride as $adminId) {
            if (!$container->hasDefinition($adminId)) {
                continue;
            }

            $definition = $container->getDefinition($adminId);
            if ($definition->hasTag('doctrine.event_subscriber')) {
                $tags = $definition->getTag('doctrine.event_subscriber');

[...]

PR #257, opened 2017-01-23
class metadata

DavidBadura:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Has tests?    | yes
| Fixed tickets | #223, #140

* [x] add class metadata
* [x] add annotation metadata driver
* [x] allow different metadata drivers
* [x] add `ignore field` annotation
* [x] remove `audited_entities` configuration
* [x] remove `global_ignore_properties`
* [x] remove hardcoded default entity manager
* [x] refactor the tests so that they are deterministic
* [x] add build symfony container test
* [x] update documentation


tests/SimpleThings/Tests/EntityAudit/Fixtures/Core/PetAudit.php
```diff
@@ -3,6 +3,7 @@
 namespace SimpleThings\EntityAudit\Tests\Fixtures\Core;
 
 use Doctrine\ORM\Mapping as ORM;
+use SimpleThings\EntityAudit\Mapping\Annotation\Auditable;
```
smoench: Unused

Soullivaneuh:
@DavidBadura So from now we can only use annotation? How about entities defined from configuration?

Example: http://symfony.com/doc/current/doctrine.html#add-mapping-information (Yaml and XML examples).

DavidBadura:
The metadata from entity audit has nothing to do with doctrine`s metadata. You can use yaml and xml to configure your entities, but (only) for auditing you need to use annotation for now.

I have created an issue for the yaml driver: https://github.com/simplethings/EntityAuditBundle/issues/262

bendavies:
@DavidBadura the listeners are no longer automatically registered after this change, making the bundle integration useless. did you realise?

DavidBadura:
@bendavies Yes, I know it. I don't have the time to finish it. We have already an issue for this problem https://github.com/simplethings/EntityAuditBundle/issues/268

bendavies:
alright, thanks!

mikeyudin:
First: Awesome work with this feature, but 1 question:  Why was global_ignore_properties removed?  This seems like a nice feature to have so devs don't have to add the ignore annotation to a ton of entities.

Issue #288, opened 2017-07-05
Problem with Detail Action


sw4n:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | ?

When trying to get audited entity details for this example route:
`/audit/viewent/AuditedEntityExample/15/29032`
I'm getting doctrine exception:

> SQLSTATE[42S22]: Column not found: 1054 Unknown column 'e.0' in 'where clause'

I've found this route while debugging:
https://github.com/simplethings/EntityAuditBundle/blob/master/src/SimpleThings/EntityAudit/Resources/config/routing.yml#L17

Which relates to this action:
https://github.com/simplethings/EntityAuditBundle/blob/master/src/SimpleThings/EntityAudit/Controller/AuditController.php#L120

On this line explode function will return bool or array, which eventually passing to this:
https://github.com/simplethings/EntityAuditBundle/blob/master/src/SimpleThings/EntityAudit/AuditReader.php#L175

And gives "0" as select field. Can you fix this?

tolry:
@sw4n a few questions

* which version are you using: caution, master is currently **not** stable/usable
* which Symfony version are you using?

sw4n:
@tolry I'm using v1.0.5 and Symfony v2.8.17

sw4n:
@tolry can you give any feedback about this?

mhnrm:
Same issue here, is there any progress in this?

I'm currently using version 1.0.6 of the AuditBundle and Symfony 2.8.25.

cirovargas:
Same here Bundle version 1.0.6, symfony 3.3.9

Ballihaut:
Hi,
we are facing to the same  
I see this issue is open since for a wild. 
Did this issue have a definive fix, we don't have identify ?

helmuthva:
same here.

helmuthva:
Same here:

2018-05-06 10:56:56] app.DEBUG: hosts from env {"hosts":"localhost:9200"} []
 --------------------- --------------------------------------------------------------------------------------------- 
  Symfony                                                                                                            
 --------------------- --------------------------------------------------------------------------------------------- 
  Version               4.0.9                                                                                        
  End of maintenance    07/2018                                                                                      
  End of life           01/2019                                                                                      
 --------------------- --------------------------------------------------------------------------------------------- 
  Kernel                                                                                                             
 --------------------- --------------------------------------------------------------------------------------------- 
[...]

PR #289, opened 2017-07-11
Issue 288

sw4n:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Has tests?    | no
| Fixed tickets | #288

### DIFF
--- a/src/SimpleThings/EntityAudit/Controller/AuditController.php
+++ b/src/SimpleThings/EntityAudit/Controller/AuditController.php
@@ -97,8 +97,7 @@ class AuditController extends Controller
      */
     public function viewEntityAction($className, $id)
     {
-        $ids = explode(',', $id);
-        $revisions = $this->getAuditReader()->findRevisions($className, $ids);
+        $revisions = $this->getAuditReader()->findRevisions($className, $id);
 
         return $this->render('SimpleThingsEntityAuditBundle:Audit:view_entity.html.twig', array(
             'id' => $id,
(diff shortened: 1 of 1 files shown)


tolry:
I do not know this part of the bundle at all, but the docblock states that `@param string $id Comma separated list of identifiers`

Anybody around who knows these bundle controllers and can tell us, if there are use cases with multiple ids? @DavidBadura @smoench @bendavies @andrewtch  ...?

Explode on a single number (or string containing a single number) should not be the problem here (https://3v4l.org/jLdJA)

As you mentioned in #288 this seems to be the faulty logic https://github.com/simplethings/EntityAuditBundle/blob/master/src/SimpleThings/EntityAudit/AuditReader.php#L175 expecting the first key in the array/hashmap to be the column name(??)

Judging from the obvious problem when using the bundle actions I would say, nobody could use them without error right now. If I do not get any objections, I will merge this.

DavidBadura:
@tolry The usecase of multiple ids is multiple primary keys ;-) You don't have in this case only one id. 

@sw4n you should check after exploding if you have only one element in the array and then pass the one element instead of the array.

tolry:
@DavidBadura ok, but explode still leaves you with numeric keys to the hashmap which cannot work properly the way the AuditReader deals with array for `$id`

tolry:
@DavidBadura just to make sure, with "multiple ids" you mean compound primary keys, right?

sw4n:
@DavidBadura I agree with @tolry in case of several primary keys, after explode the keys are stay numeric and can't be used in reader. 
Also, I must say that this pull request must be deleted or declined, because it's made from master, but not from stable version. I can make a fix, after proper deciding what to do.

nicolas-san:
Hi guys, any news on this ?
Thanks :)

Issue #286, opened 2017-06-18
Default Username null not allowed if token username not set in 1.0.5


aggayush:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | yes

- If username is not set in token storage, like in a command. Previous versions were having default username as blank string which was getting saved. Now, it throws NULL error.
- Back compatibility is breaking.

DavidBadura:
Thanks for reporting. Which version did you have before?

aggayush:
We were using 0.7 before. Now switched to 1.0.5

tolry:
Havin the same problem here, when upgrading from `^0.9` to `1.0.5` - investigating

tolry:
Hmm, this looks correct https://github.com/simplethings/EntityAuditBundle/blob/v1.0.5/src/SimpleThings/EntityAudit/EventListener/CreateSchemaListener.php#L112

But doesn't seem to work or the code is not even executed?

tolry:
Ok, the `setNotnull()` works, but this is a new(ish) thing. So previously the username would always be a string. If the username was unknown, it would be an empty string, but never null.

So one way to fix this would be to run a migration and get an `ALTER TABLE` for the username column allowing NULL. But I would still consider this a BC break.

The responsible commit is the following

https://github.com/simplethings/EntityAuditBundle/commit/4640b8ed8fcab3e40d30b6f45253285d5faeb421

In this commit, we made the way configurable, how the username is obtained (#187) and if no callable was set (or the old setUsername() was not called) it would return null

``` php
return $callable ? $callable() : null;
````

I personally do not see any merit in `null` over an empty string for a string column (so, no real FK anyway) and would like to return to the old behaviour. Any comments/objections @bendavies (committer) or @ocramius (issue creator)?

tolry:
I will prepare a PR shortly

Ocramius:
@tolry in SQL, null is a very precise concept ("absence" of a value)

tolry:
@ocramius, yes and for 2.x I might agree that this is more precise than an empty string. But forcing users to run a migration for patch level upgrades I would really like to prevent.

Ocramius:
@tolry no, indeed, for a patch version there should be no user interaction needed to upgrade.

bendavies:
(Y)

DavidBadura:
[...]

PR #291, opened 2017-07-21
Fix BC break - username should not be null

tolry:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Has tests?    | no
| Fixed tickets | #286

https://github.com/simplethings/EntityAuditBundle/commit/4640b8ed8fcab3e40d30b6f45253285d5faeb421 introduced a BC break, forcing users to run a migration on their audit tables.

This bugfix should **not** be ported to master.

### DIFF
--- a/src/SimpleThings/EntityAudit/AuditConfiguration.php
+++ b/src/SimpleThings/EntityAudit/AuditConfiguration.php
@@ -155,7 +155,7 @@ class AuditConfiguration
     {
         $callable = $this->usernameCallable;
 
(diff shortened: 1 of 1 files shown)


bendavies:
docblock is now wrong

tolry:
@bendavies yes, you are right - missed that (was a bit in a hurry, need the version for sf security fix)

tolry:
@bendavies docblock fixed directly in branch 😄

PR #293, opened 2017-07-24
fix-issue-286

tolry:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Has tests?    | no
| Fixed tickets | #286

follow up to #291 - still had the issue in a CI run, since a closure was set, but that closure returned null on CLI runs


tolry:
again, this is only relevant for 1.0, since the BC break is ok for 2.*

tolry:
thx

PR #301, opened 2017-08-31
Fix columnName

gonzaloalonsod:
$columnName = $idKeys[0]; change by $columnName = $idField;

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Has tests?    | yes
| Fixed tickets | #...

### DIFF
--- a/src/SimpleThings/EntityAudit/AuditReader.php
+++ b/src/SimpleThings/EntityAudit/AuditReader.php
@@ -234,7 +234,7 @@ class AuditReader
         foreach ($class->identifier AS $idField) {
             if (is_array($id) && count($id) > 0) {
                 $idKeys = array_keys($id);
-                $columnName = $idKeys[0];
+                $columnName = $idField;
             } else if (isset($class->fieldMappings[$idField])) {
                 $columnName = $class->fieldMappings[$idField]['columnName'];
             } elseif (isset($class->associationMappings[$idField])) {


tolry:
@gonzakpo could you elaborate which problem you had that this PR fixes? Your fix seems to break lots of unit tests, see travis logs (e.g. https://travis-ci.org/simplethings/EntityAuditBundle/jobs/270498009) - so this seems not to be universally applicable.

gonzakpo:
before it generated: WHERE e.rev <=? AND e.0 =?
and with e.0 it does not work for me.
then now it generates me: WHERE e.rev <=? AND e.id =?
and if it works for me.
What I can be doing wrong?

tolry:
@gonzakpo do you have the code accessible somewhere or can you post your entity definitions?

Webonaute:
I have the same issue, PHP7.1 with symfony 3.2.

kevinpapst:
Same issue here with PHP 7.1.12, Symfony 3.4.3, DoctrineORM 2.6.0, Doctrine-Bundle 1.8.1
Happens when I try to view the details of an audit entry. Exception:
```
An exception occurred while executing 'SELECT e.revtype, e.id AS "id", e.name AS "name", e.comment AS "comment", e.visible AS "visible", e.company AS "company" FROM xxx_customers_audit e WHERE e.rev <= ? AND e.0 = ? ORDER BY e.rev DESC' with params ["2", "2"]:
```
The fix by @gonzakpo seems to work in my scenario.

tolry:
Hmm, I guess this might be an incompatibility with Doctrine 2.6 - anyone here having this issue with 2.5 or lower?

Probably noone around with enough internals knowledge to give a hint what might be the issue here? As I sait: the proposed fix seems to break a few tests and probably breaks production code ... hmm

kevinpapst:
Its good that it breaks unit tests, a sign for a well tested bundle! But if these test cover a bug, that might be a false-positive then ;-) at least with the provided software package versions. 
Let me know if I can be of any help in investigating. Unfortunately I can neither share my repo (private) nor downgrade the Doctrine packages. 

And as you said: do not simply apply the patch to the core , as it will very likely break production code for other users out-there. I just wanted to make sure that you are aware of the problem and keep an eye on it.

PR #324, opened 2018-04-11
Update `AuditedCollection` in order to respect `Collection` interface

### LINKED ISSUE #308: `AuditCollection` methods intended to return a collection of entities should not return an array
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| Used Version     | 1.0.6

`AuditedCollection` methods intended to return a collection of entities (`partition()`, `filter()`, `map()`) should not return an array, but an object implementing `Collection` interface with the resulting elements.
This behavior must be the same which is implemented by `PersistentCollection`, otherwise an error occurs when the audited entity uses any of these methods.

Take a look at this example:
```php
public function __construct()
{
    $this->users = new ArrayCollection();
}

public function getActiveUsers(): ?ArrayCollection
{
    $activeUsers = $this->users->filter(function (User $user) {
        return $user->isActive();
    });

    // `$activeUsers` is an array when the entity is loaded by the audit bundle, so
    // `isEmpty()` call will result in "Call to a member function isEmpty() on array"
    if ($activeUsers->isEmpty()) {
        return null;
    }

    return $activeUsers;
}

### PR DESCRIPTION
**Pull request by phansys**
## Subject
Update `AuditedCollection` in order to respect `Collection` interface.

I am targeting this branch, because these changes attempt to respect BC.

Closes #308.

## Changelog

```markdown
### Fixed
- `AuditedCollection` methods in order to respect `Collection` interface.
```

### DIFF
--- a/src/Collection/AuditedCollection.php
+++ b/src/Collection/AuditedCollection.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace SimpleThings\EntityAudit\Collection;
 
+use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use SimpleThings\EntityAudit\AuditConfiguration;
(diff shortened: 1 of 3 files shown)


phansys:
Now, this PR has 2 commits. First (835303a) just adds creates new instances of `ArrayCollection` in specific methods; while the second (258089e) decorates `ArrayCollection` for the whole class (which seems more consistent IMO).
Please share your concerns in order to decide which one to pick.
Thank you in advance.

phansys:
Since I think you don't have enough time to review or make a deep analysis on this fix, I chose the 2nd option and squashed my commits.
From my point of view, It is ready to merge.

VincentLanglet (approved):

wbloszyk (approved):

VincentLanglet:
Thanks @phansys

PR #327, opened 2018-06-25
Global Ignores, New Annotations, Comparators, Schema Creation based on annotations working

brainwood:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Has tests?    | yes/no
| Fixed tickets | unknown

This pull requests introduces global ignores back into the configuration, as well as adding two new annotations to deal with inherited attributes or overriding ignores for specific cases. It also  It also introduces the concept of Comparators which allows users to define services that can determine if a specific change is significant enough to trigger a revision to be created, for example a float value may have some insignificant change that still set off the EntityChangeSet (0.00 -> 0) a comparator can be created to compare the values and remove this from the ChangeSet that is considered by the LogRevision listener.


brainwood:
This is the same branch on my forks travis-ci https://travis-ci.org/brainwood/EntityAuditBundle/builds/396498140

PR #331, opened 2018-09-19
Fix unknown column 'e.0' in 'where clause' on AuditController

torreytsui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Has tests?    | no
| Fixed tickets | #288, #289

`\SimpleThings\EntityAudit\AuditReader::find` expect a string `$id` or an array of id field / value pair.

Example:

```php
$id = 1
```

or 

```php
$id = [
  'id' => 1,
];
```

AuditController use `explode(',', $id);` caused an unexpected 'e.0' where clause generated:

```
An exception occurred while executing 'SELECT e.revtype, e.id AS `id`, e.plus_one AS `plusOne`, e.respondent_id, e.wedding_event_id FROM wedding_event_invite_audit e WHERE e.rev <= ? AND e.0 = ? ORDER BY e.rev DESC' with params ["108", "86"]:

SQLSTATE[42S22]: Column not found: 1054 Unknown column 'e.0' in 'where clause'
```

I believe this controller is most likely for demo purpose only. But it is still worth fixing.

### DIFF
--- a/src/SimpleThings/EntityAudit/Controller/AuditController.php
+++ b/src/SimpleThings/EntityAudit/Controller/AuditController.php
@@ -97,8 +97,7 @@ class AuditController extends Controller
      */
     public function viewEntityAction($className, $id)
     {
-        $ids = explode(',', $id);
-        $revisions = $this->getAuditReader()->findRevisions($className, $ids);
+        $revisions = $this->getAuditReader()->findRevisions($className, $id);
 
         return $this->render('SimpleThingsEntityAuditBundle:Audit:view_entity.html.twig', array(
             'id' => $id,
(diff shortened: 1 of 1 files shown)


tolry:
related to #288

torreytsui:
Also related to #289. Sorry I should have done a proper search.

The only difference is that 289 is targeting master. This PR is targeting 1.0 branch.

tolry:
No problem - just wanted to quickly drop the reference, so this isn't missed.

As you might have noticed, there are no developers currently active in this project - so it might be a while, before this is merged. If you need this fix in the near future, I would recommend using your own fork for now.

PR #335, opened 2019-02-27
Fix memory leak in LogRevisionsListener.php on post flush

Avoran:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Has tests?    | no

When you update several thousand entities in small batches, the extraUpdates array keeps all the entities. This is unnecessary and a memory leak. By emptying it after the update data is written it removes said memory leak.


Avoran:
First test(continuousphp) was merging into master. How can I rerun this test for the current pull-request with a merge into 1.0?

martin-georgiev:
This PR duplicates #333

PR #347, opened 2021-02-07
Use organisation similar to other sonata package

VincentLanglet:
Pedantic
Related to https://github.com/sonata-project/dev-kit/issues/1405


.php_cs
```diff
@@ -1,44 +0,0 @@
-<?php
-
-$header = <<<EOF
```
wbloszyk, line 3: IMO We should keep this file. We are keep `SimpleThings\EntityAuditBundle` so header should be keep. We can add note about we are maintainer since xxx. 

WDYT? @sonata-project/contributors

VincentLanglet: All these files will be autogenerated by dev-kit, so it's easier this way.

I prefer to add a custom note in the README instead.

wbloszyk:
@VincentLanglet Can you create `Move CI to github actions` commit as seperate PR? This should allow use acions for other changes.

phpunit.xml.dist
```diff
@@ -1,9 +1,35 @@
...
+         colors="true"
+         failOnWarning="true"
+         failOnRisky="true"
+         stopOnFailure="false"
+         bootstrap="tests/bootstrap.php"
+>
     <testsuites>
-        <testsuite name="SimpleThings EntityAudit">
-            <directory>./tests/SimpleThings/Tests/EntityAudit</directory>
+        <testsuite name="SonataDoctrineOrmAdminBundle Test Suite">
```
wbloszyk, line 18: Wrong naming

VincentLanglet: Fixed

VincentLanglet:
Tests are passing https://github.com/VincentLanglet/EntityAuditBundle/pull/1

For the lint actions, it's another story ^^', so I prefer to split.

OskarStark:
Looks good to me, shall we merge?

.php_cs
wbloszyk: I did not think about it. We can do it too.

.php_cs.dist
```diff
@@ -0,0 +1,76 @@
...
+/*
+ * DO NOT EDIT THIS FILE!
+ *
+ * It's auto-generated by sonata-project/dev-kit package.
+ */
+
+$header = <<<'HEADER'
+This file is part of the Sonata Project package.
+
+(c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
```
core23, line 12: This is okay? The original package was not created by thomas.

VincentLanglet: For what I understand, it means that it's inside the `sonata-project` organisation, which is true now.

wbloszyk (approved):
IMO It is almost RTM.

Last thing is decide wha to do with files header: 
https://github.com/sonata-project/EntityAuditBundle/pull/347/files#r571652696
https://github.com/sonata-project/EntityAuditBundle/pull/347/files#r571875554

IMO We should keed old version (Info about license can be replaced).

WDYT? @sonata-project/contributors

VincentLanglet:
> IMO We should keed old version (Info about license can be replaced).

It's easier to change them:
- The dev-kit coding standard won't be usable if we don't change them
- The dev-kit won't be able to update automatically file and to use our template, if we don't change them

If you look at our  header:
> This file is part of the Sonata Project package.
> (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>

This is true, the entityAuditBundle is now reference as a sonata project repository
https://github.com/sonata-project/EntityAuditBundle

> For the full copyright and license information, please view the LICENSE
file that was distributed with this source code.

We can have a custom LICENSE if necessary, and add some explanations in the README.

OskarStark:
I am fine with changing the header 👍🏻

OskarStark:
cc @tolry wdyt about the Header change?

tolry:
since I haven't contributed almost any code to the repository I will delegate this to @beberlei since he was the initial author and probably the best one to comment on this :slightly_smiling_face:

tolry:
no objection from him :)

OskarStark:
Thanks for the feedback 👍

PR #323, opened 2018-04-10
[bugfix] Use long array syntax at `IssueTest::testIssue318()`

phansys:
|Q            |A   |
|---          |--- |
|Branch       |1.0 |
|Bug fix?     |yes |
|New feature? |no  |
|BC breaks?   |no  |
|Deprecations?|no  |
|Tests pass?  |yes |
|Fixed tickets|#320|
|License      |n/a |
|Doc PR       |n/a |

PR #350, opened 2021-02-12
Allow php8

VincentLanglet:


phansys (approved):

Yozhef:
@VincentLanglet could you release a new release to unlock work with updates to the next sonata project

wbloszyk:
@Yozhef for now we are working on https://github.com/sonata-project/dev-kit/issues/1405#issuecomment-781373972 and I do not think we wants to make release before these works. We have to decide about do we wanna provide entity audit also as standalone library. Additionaly we have to check code with commits and update changelog.

For now You can use `dev-1.x` version.

VincentLanglet:
> @Yozhef for now we are working on [sonata-project/dev-kit#1405 (comment)](https://github.com/sonata-project/dev-kit/issues/1405#issuecomment-781373972) and I do not think we wants to make release before these works. We have to decide about do we wanna provide entity audit also as standalone library. Additionaly we have to check code with commits and update changelog.

I disagree, we don't have to wait for this to release a new version. I think we can tag a new one now or soon.

wbloszyk:
Release mean code is stable. Previous system was working on merge. In some cases code are appered and desapper in next tag. We must check it and understand it correct or it is mistake. So I do not think tag it is good idea.

VincentLanglet:
The branch 1.x is based on the branch 1.0.
The code of the branch 1.0.9 (the last release) was working so I don't see why it wouldn't work now.

wbloszyk:
Look at https://github.com/sonata-project/EntityAuditBundle/pull/323#issuecomment-380098558

This is comment for fix 5725eca

VincentLanglet:
> Look at [#323 (comment)](https://github.com/sonata-project/EntityAuditBundle/pull/323#issuecomment-380098558)
> 
> This is comment for fix [5725eca](https://github.com/sonata-project/EntityAuditBundle/commit/5725ecab2d132288842d9a075d1a8662520829c5)

It's not because the test suite was broken that the bundle wasn't working. It was used a lot.
And the test suite is not broken anymore.
Even if it was broken, if we release a new version, it wouldn't be worse than the 1.0.9 one, so I see no reason to not doing it.

The only question should be, do we have introduce a bug since 1.0.9 ? I think not.

wbloszyk:
@Yozhef 
https://github.com/sonata-project/EntityAuditBundle/releases/tag/1.1.0

PR #355, opened 2021-02-16
Allow to use other connection and entity manager than "default"

phansys:
## Subject
Allow to use other connection and entity manager than "default".

I am targeting this branch, because these changes respect BC.

Closes #140.

## Changelog

```markdown
### Added
- `connection` configuration node in order to use a different connection than "default";
- `entity_manager` configuration node in order to use a different entity manager than "default".
```

PR #363, opened 2021-02-18
Fix connection for doctrine event tag

wbloszyk:
## Subject

After #355 I got some errors like:
```
The Doctrine connection "%simplethings.entityaudit.connection%" referenced   
!!    in service "simplethings_entityaudit.log_revisions_listener" does not exist  
!!    . Available connections names: "default".   
```

It look like doctrine event tag will not convert parameter name into value. This PR will replace this parameter name into value.

I am targeting this branch, because this change respect BC.

Changelog is not needed becouse we do not release #355 yet.

### DIFF
--- a/src/DependencyInjection/SimpleThingsEntityAuditExtension.php
+++ b/src/DependencyInjection/SimpleThingsEntityAuditExtension.php
@@ -49,5 +49,26 @@ class SimpleThingsEntityAuditExtension extends Extension
                 $container->setAlias('simplethings_entityaudit.'.$key, $service);
             }
         }
+
+        $this->fixConnectionForDoctrineTag($container, [
+            'simplethings_entityaudit.log_revisions_listener',
+            'simplethings_entityaudit.create_schema_listener',
+        ]);
+    }
+
+    private function fixConnectionForDoctrineTag(ContainerBuilder $container, array $definitionNames): void
+    {
+        foreach ($definitionNames as $definitionName) {
+            $definition = $container->getDefinition($definitionName);
+            $definition->clearTag('doctrine.event_subscriber');
+
+            foreach ($definition->getTag('doctrine.event_subscriber') as $id => $attributes) {
+                if (isset($attributes['connection']) && $container->hasParameter($attributes['connection'])) {
+                    $attributes['connection'] = (string) $container->getParameter('simplethings.entityaudit.connection');
+                }
+
+                $definition->addTag('doctrine.event_subscriber', $attributes);
+            }
+        }
     }
 }
--- a/src/DependencyInjection/SimpleThingsEntityAuditExtension.php
+++ b/src/DependencyInjection/SimpleThingsEntityAuditExtension.php
@@ -50,23 +50,23 @@ class SimpleThingsEntityAuditExtension extends Extension
             }
         }
 
-        $this->fixConnectionForDoctrineTag($container, [
+        $this->fixParametersFromDoctrineEventSubscriberTag($container, [
             'simplethings_entityaudit.log_revisions_listener',
             'simplethings_entityaudit.create_schema_listener',
         ]);
     }
 
-    private function fixConnectionForDoctrineTag(ContainerBuilder $container, array $definitionNames): void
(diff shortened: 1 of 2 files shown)


wbloszyk:
@phansys 
I check in my project and it is working. Anyway I can not fix tests. Can you look at this?

phansys:
> @phansys
> I check in my project and it is working. Anyway I can not fix tests. Can you look at this?

Sure, I'll try to check this later.

phansys:
I think this is working, at least in the test environment: wbloszyk/EntityAuditBundle#1.
Please, confirm if you can check these changes in a working project.

wbloszyk:
> I think this is working, at least in the test environment: [wbloszyk#1](https://github.com/wbloszyk/EntityAuditBundle/pull/1).
> Please, confirm if you can check these changes in a working project.

@phansys 
Parameter will be always the same `simplethings.entityaudit.connection`. We do not have to check it, becouse it is set in `load()`.

src/DependencyInjection/SimpleThingsEntityAuditExtension.php
```diff
@@ -49,5 +49,26 @@ public function load(array $configs, ContainerBuilder $container): void
...
+    }
+
+    private function fixParametersFromDoctrineEventSubscriberTag(ContainerBuilder $container, array $definitionNames): void
+    {
+        foreach ($definitionNames as $definitionName) {
+            $definition = $container->getDefinition($definitionName);
+            $tags = $definition->getTag('doctrine.event_subscriber');
+            $definition->clearTag('doctrine.event_subscriber');
+
+            foreach ($tags as $id => $attributes) {
```
phansys, line 66:
```suggestion
            foreach ($tags as $attributes) {
```

phansys (changes requested):

phansys (approved):

src/DependencyInjection/SimpleThingsEntityAuditExtension.php
```diff
@@ -49,5 +49,26 @@ public function load(array $configs, ContainerBuilder $container): void
...
+    private function fixParametersFromDoctrineEventSubscriberTag(ContainerBuilder $container, array $definitionNames): void
+    {
+        foreach ($definitionNames as $definitionName) {
+            $definition = $container->getDefinition($definitionName);
+            $tags = $definition->getTag('doctrine.event_subscriber');
+            $definition->clearTag('doctrine.event_subscriber');
+
+            foreach ($tags as $attributes) {
+                if (isset($attributes['connection'])) {
+                    $attributes['connection'] = (string) $container->getParameter('simplethings.entityaudit.connection');
```
phansys, line 68: IMO, if we harcode the parameter name here, we are making less flexible the replacing mechanism.
If in the future the parameter name used in these tags is renamed, we must replace this occurrence too.
I also think that the final users could not override these services if they need to use a different value.

wbloszyk: We set this `simplethings.entityaudit.connection` parameter in the same `SimpleThingsEntiyAuditExtension`, so I think we do not have to afriaid about rename. Additionaly if users override this service then connection name must be set directly, oherwise expection will be throw. 

We can also move it to the compiler pass. Then we can check whole services tagged by `doctrine.event_listener` and replace `simplethings.entityaudit.connection` parameter in `connetion` to propertly value.

wbloszyk: https://github.com/symfony/symfony/pull/40262
WDYT? @sonata-project/contributors

VincentLanglet: I'm not sure to understand, should we wait for https://github.com/symfony/symfony/pull/40262 or can we merge this ?

wbloszyk: We can merge this now. This code allow use param only for ower services.we can remove it when symfony accept it and we bump doctrine-bridge.

VincentLanglet (approved):

PR #365, opened 2021-02-20
Improve routing

wbloszyk:
## Subject

Add xml routing o be consider in other sonata bundles routing.

I am targeting this branch, because change respect BC.

## Changelog

```markdown
### Added
- Added routing in xml: `src\Resources\config\routing\audit.xml`
```

### DIFF
--- a/src/Resources/config/routing.yml
+++ b/src/Resources/config/routing.yml
@@ -1,19 +1,30 @@
 simple_things_entity_audit_home:
-    path: /{page<\d+>?1}
+    path: /{page}
     controller: SimpleThings\EntityAudit\Action\IndexAction
+    defaults:
+        page: 1
+    requirements:
+        page: \d+
 
 simple_things_entity_audit_viewrevision:
-    path: /viewrev/{rev<\d+>}
+    path: /viewrev/{rev}
     controller: SimpleThings\EntityAudit\Action\ViewRevisionAction
+    requirements:
+        rev: \d+
 
 simple_things_entity_audit_viewentity_detail:
-    path: /viewent/{className}/{id}/{rev<\d+>}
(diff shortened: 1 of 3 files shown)


wbloszyk:
@OskarStark @VincentLanglet 
Can we allow adding defaults and requiurements in path for xml files?

phansys:
For semantic reasons, I think we should keep the these XML nodes separated. Configuration in reusable bundles should be as expressive as possible.

VincentLanglet (approved):

src/Resources/config/routing.yml
```diff
@@ -1,30 +1,19 @@
 simple_things_entity_audit_home:
-    path: /{page}
+    path: /{page<\d+>?1}
```
phansys, line 2: IMO, these changes should be reverted.

wbloszyk: We do not have to but I should add next_maroj note about remove this file.

wbloszyk: Reverted. Also I add NEXT_MAJOR to remove this file.

src/Resources/config/routing.yml
```diff
@@ -1,3 +1,4 @@
+# NEXT_MAJOR: remove this file and add upgrade note
```
phansys, line 1: Shouldn't we update the documentation at [`README.md`](https://github.com/sonata-project/EntityAuditBundle/tree/d68de63687b49d4859ae233ab90bc40e7b089add#appconfigroutingyml) in order to use the routes from the XML file?

wbloszyk: Done.

VincentLanglet (approved):

phansys (approved):

PR #382, opened 2021-04-04
Improve typehint

VincentLanglet:
## Subject

I am targeting this branch, because BC.

The `$revision` param is often restrict to `int` but could work with string too. (and SonataAdmin is using `int|string`).
The `$id` seems to only work with `int|string|array` if I read the code.

I also add some typehint for constructor or Exception/Object we're using.

This will fix some errors from https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1386/checks?check_run_id=2265615169

## Changelog

```markdown
### Added
- Allow `$revisions` param to be a string in `AuditManager` methods.

### Deprecated
- Passing another value than 'string', 'integer', 'smallint', 'bigint' or 'guid' for the `revision_id_field_type` value.
```

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -18,6 +18,14 @@ use Symfony\Component\Config\Definition\ConfigurationInterface;
 
 class Configuration implements ConfigurationInterface
 {
+    private const ALLOWED_REVISION_ID_FIELD_TYPE = [
+        'string',
+        'integer',
+        'smallint',
+        'bigint',
+        'guid',
+    ];
+
     public function getConfigTreeBuilder()
     {
         $builder = new TreeBuilder('simple_things_entity_audit');
@@ -36,7 +44,25 @@ class Configuration implements ConfigurationInterface
                 ->scalarNode('revision_field_name')->defaultValue('rev')->end()
                 ->scalarNode('revision_type_field_name')->defaultValue('revtype')->end()
                 ->scalarNode('revision_table_name')->defaultValue('revisions')->end()
-                ->scalarNode('revision_id_field_type')->defaultValue('integer')->end()
+                ->scalarNode('revision_id_field_type')
+                    ->defaultValue('integer')
+                    // NEXT_MAJOR: Use validate() instead.
+                    ->beforeNormalization()
+                        ->always(static function ($value) {
+                            if (null !== $value && !\in_array($value, self::ALLOWED_REVISION_ID_FIELD_TYPE, true)) {
+                                @trigger_error(sprintf(
+                                    'The value "%s" for the "revision_id_field_type" is deprecated'
+                                    .' since sonata-project/entity-audit-bundle 1.3 and will throw an error in version 2.0.'
+                                    .' You must pass one of the following values: "%s".',
+                                    $value,
+                                    implode('", "', self::ALLOWED_REVISION_ID_FIELD_TYPE)
+                                ), \E_USER_DEPRECATED);
+                            }
+
+                            return $value;
+                        })
+                    ->end()
+                ->end()
                 ->arrayNode('service')
                     ->addDefaultsIfNotSet()
                     ->children()
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -46,7 +46,7 @@ class Configuration implements ConfigurationInterface
                 ->scalarNode('revision_table_name')->defaultValue('revisions')->end()
                 ->scalarNode('revision_id_field_type')
                     ->defaultValue('integer')
-                    // NEXT_MAJOR: Use validate() instead.
+                    // NEXT_MAJOR: Use enumNode() instead.
                     ->beforeNormalization()
                         ->always(static function ($value) {
                             if (null !== $value && !\in_array($value, self::ALLOWED_REVISION_ID_FIELD_TYPE, true)) {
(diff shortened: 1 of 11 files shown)


phansys (commented):
Checking the configuration, the revision id seems to accept anything:
https://github.com/sonata-project/EntityAuditBundle/blob/a800c2a0781b162e3c9636cf593fff55483bbe18/src/DependencyInjection/Configuration.php#L39
Maybe we could restrict the config to "integer" and "string".

VincentLanglet:
> Maybe we could restrict the config to "integer" and "string".

It would be a BC-break no ? I think we will be able to restrict only for the next major...

phansys:
>It would be a BC-break no ? I think we will be able to restrict only for the next major...

Yes. We should deprecate the types that we consider invalid before.

VincentLanglet:
@phansys There is the following code:
```
$revisionsTable->addColumn('id', $this->config->getRevisionIdFieldType(), [
            'autoincrement' => true,
        ]);
```
Would it work with string ?

phansys:
>Would it work with string ?

I think the `autoincrement` option is ignored for other types.

src/DependencyInjection/Configuration.php
```diff
@@ -36,7 +44,25 @@ public function getConfigTreeBuilder()
...
+                    // NEXT_MAJOR: Use validate() instead.
+                    ->beforeNormalization()
+                        ->always(static function ($value) {
+                            if (null !== $value && !\in_array($value, self::ALLOWED_REVISION_ID_FIELD_TYPE, true)) {
+                                @trigger_error(sprintf(
+                                    'The value "%s" for the "revision_id_field_type" is deprecated'
+                                    .' since sonata-project/entity-audit-bundle 1.3 and will throw an error in version 2.0.'
+                                    .' You must pass one of the following value: %s.',
+                                    $value,
+                                    implode(', ', self::ALLOWED_REVISION_ID_FIELD_TYPE)
```
phansys, line 58:
```suggestion
                                    implode(', ', self::ALLOWED_REVISION_ID_FIELD_TYPE)
```
```suggestion
                                    implode('", "', self::ALLOWED_REVISION_ID_FIELD_TYPE)
```

src/DependencyInjection/Configuration.php
```diff
@@ -36,7 +44,25 @@ public function getConfigTreeBuilder()
...
+                ->scalarNode('revision_id_field_type')
+                    ->defaultValue('integer')
+                    // NEXT_MAJOR: Use validate() instead.
+                    ->beforeNormalization()
+                        ->always(static function ($value) {
+                            if (null !== $value && !\in_array($value, self::ALLOWED_REVISION_ID_FIELD_TYPE, true)) {
+                                @trigger_error(sprintf(
+                                    'The value "%s" for the "revision_id_field_type" is deprecated'
+                                    .' since sonata-project/entity-audit-bundle 1.3 and will throw an error in version 2.0.'
+                                    .' You must pass one of the following value: %s.',
```
phansys, line 56:
```suggestion
                                    .' You must pass one of the following values: "%s".',
```

src/DependencyInjection/Configuration.php
```diff
@@ -36,7 +44,25 @@ public function getConfigTreeBuilder()
...
-                ->scalarNode('revision_id_field_type')->defaultValue('integer')->end()
+                ->scalarNode('revision_id_field_type')
+                    ->defaultValue('integer')
+                    // NEXT_MAJOR: Use validate() instead.
+                    ->beforeNormalization()
+                        ->always(static function ($value) {
+                            if (null !== $value && !\in_array($value, self::ALLOWED_REVISION_ID_FIELD_TYPE, true)) {
+                                @trigger_error(sprintf(
+                                    'The value "%s" for the "revision_id_field_type" is deprecated'
+                                    .' since sonata-project/entity-audit-bundle 1.3 and will throw an error in version 2.0.'
```
phansys, line 55:
```suggestion
                                    .' since sonata-project/entity-audit-bundle 1.x and will throw an error in version 2.0.'
```

phansys (changes requested):

src/DependencyInjection/Configuration.php
VincentLanglet: I'm the only PR opened, and since it's a deprecation, it will be a minor. I win some time ^^'

phansys (approved):

src/DependencyInjection/Configuration.php
```diff
@@ -36,7 +44,25 @@ public function getConfigTreeBuilder()
                 ->scalarNode('revision_field_name')->defaultValue('rev')->end()
                 ->scalarNode('revision_type_field_name')->defaultValue('revtype')->end()
                 ->scalarNode('revision_table_name')->defaultValue('revisions')->end()
-                ->scalarNode('revision_id_field_type')->defaultValue('integer')->end()
+                ->scalarNode('revision_id_field_type')
+                    ->defaultValue('integer')
+                    // NEXT_MAJOR: Use validate() instead.
```
franmomu, line 49: Use [enumNode](https://symfony.com/doc/4.4/components/config/definition.html#enum-nodes) instead?

src/Exception/AuditException.php
```diff
@@ -15,13 +15,22 @@
...
+     */
     protected $id;
 
+    /**
+     * @var int|string|null
+     */
     protected $revision;
 
-    public function __construct($className, $id, $revision)
+    public function __construct(?string $className, ?array $id, $revision)
```
franmomu, line 33:
```suggestion
    /**
     * @param int|string|null $revision
     */
    public function __construct(?string $className, ?array $id, $revision)
```

franmomu (dismissed):

PR #394, opened 2021-04-21
add two new options (ignore specific column & convert enum to string)

michealmouner:
## Subject

I am targeting this branch, because I need to add two new features to the bundle
so, the bundle has an option for ignoring a column from all tables ex: `updatedAt, createdAt`
but sometimes we do need to track some specific columns for a specific table and not in general
I added this option

```
    global_ignore_columns: # old one still working
        - created_at
        - updated_at

    table_ignore_columns: # new one
        - TABLE_NAME.COLUMN_NAME
        - TABLE_NAME.COLUMN_NAME
```

also, I added a boolean option to convert EnumToString incase you are using it in the DB (it causes problems)
```
```

## Changelog

```markdown
### Added
two new options:
- table_ignore_columns --> allow to ignore table for specific column

### Fixed
src/SimpleThings/EntityAudit/Utils/SimpleDiff.php:33 --> fix undefined variable

```


michealmouner:
#395

PR #244, opened 2017-01-09
Rename all global_ignore_columns to global_ignore_properties

jenyak:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Has tests?    | no

This PR fix error
```
[RuntimeException]                                                                                                                                                                                                       
  An error occurred when executing the "'cache:clear --no-warmup'" command:                                                                                                                                                
                                                                                                                                         
    [Symfony\Component\DependencyInjection\Exception\ParameterNotFoundException]                                                                                                                                           
                                                                                                                                                                                                                           
    The service "simplethings_entityaudit.config" has a dependency on a non-existent parameter "simplethings.entityaudit.global_ignore_properties". Did you mean one of these: "simplethings.entityaudit.global_ignore_co  
  lumns", "simplethings.entityaudit.table_prefix"?    
```

Because not all global_ignore_columns was renamed after #244

PR #395, opened 2021-04-21
Add two new options (ignore specific column & convert enum to string)

michealmouner:
## Subject

I am targeting this branch, because I need to add two new features to the bundle
so, the bundle has an option for ignoring a column from all tables ex: `updatedAt, createdAt`
but sometimes we do need to track some specific columns for a specific table and not in general
I added this option

```
global_ignore_columns: # old one still working
    - created_at
    - updated_at

entity_ignored_properties: # new one
    App\Entity\EntityOne:
        - propertyOne
        - propertyTwo
    App\Entity\EntityTwo:
        - propertyOne
```

also, I added a boolean option to convert EnumToString incase you are using it in the DB (it causes problems)
```
```

example of created ENUM migration
```
$this->addSql('ALTER TABLE table_name ADD enum_column ENUM(\'option_1\', \'option_2\', \'option_3\') DEFAULT \'option_3\' NOT NULL');
$this->addSql('ALTER TABLE table_name ADD enum_column VARCHAR(255) DEFAULT \'option_3\'');
```

## Changelog

```markdown
### Added
two new options:
- table_ignore_columns --> allow to ignore table for specific column

### Fixed
src/SimpleThings/EntityAudit/Utils/SimpleDiff.php:26 --> fix undefined variable

```


src/AuditConfiguration.php
```diff
@@ -172,4 +174,29 @@ public function getRevisionIdFieldType()
     {
         return $this->revisionIdFieldType;
     }
+
+    public function getTableIgnoreColumns()
```
VincentLanglet, line 178: You can add return type hint to all these new functions.

src/AuditManager.php
```diff
@@ -49,9 +50,9 @@ public function createAuditReader(EntityManager $em)
         return new AuditReader($em, $this->config, $this->metadataFactory);
     }
 
-    public function registerEvents(EventManager $evm): void
+    public function registerEvents(EventManager $evm, EntityManagerInterface $em): void
```
VincentLanglet, line 53: This is a BC-break.

tests/Fixtures/Core/ProfileAudit.php
```diff
@@ -32,6 +32,11 @@ class ProfileAudit
      */
     private $biography;
 
+    /**
+     * @ORM\Column(type="string")
+     */
+    private $ignoreme;
```
VincentLanglet, line 38: ignoreMe

phansys (commented):
Great job @michealmouner!

I have some concerns about these additions, since I think we should move to model properties instead of keep the column approach.
I'd like to discuss about this topic before with @sonata-project/contributors. WDYT?
See #213, #244.

michealmouner:
> Great job @michealmouner!
> 
> I have some concerns about these additions, since I think we should move to model properties instead of keep the column approach.
> I'd like to discuss about this topic before with @sonata-project/contributors. WDYT?
> See #213, #244.

I'm checking this right now .. I agree & should be fine I think 🤞

src/AuditConfiguration.php
```diff
@@ -19,29 +19,28 @@ class AuditConfiguration
...
     private $revisionIdFieldType = 'integer';
     private $usernameCallable;
+    private $convertEnumToString = false;
 
     /**
-     * @return AuditConfiguration
+     * @param string[] $classes
      */
-    public static function forEntities(array $classes)
+    public static function forEntities(array $classes): self
```
phansys, line 35: We can not make changes like this for method signatures in minor versions, since they break BC.

phansys (changes requested):

src/AuditConfiguration.php
michealmouner: I added hints only for now

src/AuditManager.php
```diff
@@ -44,14 +48,19 @@ public function getConfiguration()
         return $this->config;
     }
 
-    public function createAuditReader(EntityManager $em)
+    public function createAuditReader(?EntityManager $em = null)
```
phansys, line 51: This change breaks BC.

src/EventListener/CreateSchemaListener.php
```diff
@@ -74,9 +83,33 @@ public function postGenerateSchemaTable(GenerateSchemaTableEventArgs $eventArgs)
         );
 
         foreach ($entityTable->getColumns() as $column) {
+            $columnTypeName = $column->getType()->getName();
+            $columnArrayOptions = $column->toArray();
+
+            //ignore specific fields for subclasses in-case of using discriminator column
```
phansys, line 89:
```suggestion
            // Ignore specific fields for subclasses in-case of using discriminator column.
```

src/EventListener/CreateSchemaListener.php
```diff
@@ -74,9 +83,33 @@ public function postGenerateSchemaTable(GenerateSchemaTableEventArgs $eventArgs)
...
+            foreach ($cm->subClasses as $subClass) {
+                if ($cm->hasField($column->getName()) || $cm->hasAssociation($column->getName())) {
+                    if ($this->config->isEntityIgnoredProperty($subClass, $cm->getFieldForColumn($column->getName()))) {
+                        continue 2;
+                    }
+                }
+            }
+
+            //ignore specific fields for table
```
phansys, line 98:
```suggestion
            // Ignore specific fields for table.
```

src/EventListener/CreateSchemaListener.php
```diff
@@ -74,9 +83,33 @@ public function postGenerateSchemaTable(GenerateSchemaTableEventArgs $eventArgs)
...
+            if (empty($cm->discriminatorColumn) && $this->config->isEntityIgnoredProperty($cm->getName(), $cm->getFieldForColumn($column->getName()))) {
+                continue;
+            }
+
+            // change Enum type to String
```
phansys, line 103:
```suggestion
            // Change Enum type to string.
```

src/EventListener/CreateSchemaListener.php
```diff
@@ -74,9 +83,33 @@ public function postGenerateSchemaTable(GenerateSchemaTableEventArgs $eventArgs)
...
+            $sqlString = $column->getType()->getSQLDeclaration($columnArrayOptions, $this->em->getConnection()->getDatabasePlatform());
+            if ($this->config->convertEnumToString() && false !== strpos($sqlString, 'ENUM')) {
```
phansys, line 105: What are the possible strings containing "ENUM" as substring?

src/EventListener/LogRevisionsListener.php
```diff
@@ -134,6 +134,11 @@ public function postFlush(PostFlushEventArgs $eventArgs): void
                     }
                 }
 
+                //ignore specific fields for table
```
phansys, line 137:
```suggestion
                // Ignore specific fields for table.
```

src/EventListener/LogRevisionsListener.php
```diff
@@ -226,6 +231,16 @@ public function postUpdate(LifecycleEventArgs $eventArgs): void
             }
         }
 
+        // Make sure that ignored columns for table are removed from the changeset
```
phansys, line 234:
```suggestion
        // Make sure that ignored columns for table are removed from the changeset.
```

src/EventListener/LogRevisionsListener.php
```diff
@@ -370,6 +385,11 @@ private function getInsertRevisionSQL($class)
                     continue;
                 }
 
+                //ignore specific fields for table
```
phansys, line 388:
```suggestion
                // Ignore specific fields for table.
```

src/EventListener/LogRevisionsListener.php
```diff
@@ -444,6 +470,11 @@ private function saveRevisionEntityData($class, $entityData, $revType): void
                 continue;
             }
 
+            //ignore specific fields for table
```
phansys, line 473:
```suggestion
            // Ignore specific fields for table.
```

src/AuditConfiguration.php
```diff
@@ -163,13 +210,50 @@ public function getUsernameCallable()
...
+    {
+        return \array_key_exists($entity, $this->getEntityIgnoredProperties()) && \in_array($propertyName, $this->getEntityIgnoredProperties()[$entity], true);
+    }
+
+    public function setConvertEnumToString(bool $convertEnum): void
+    {
+        $this->convertEnumToString = $convertEnum;
+    }
+
+    public function convertEnumToString(): bool
```
phansys, line 255: This name doesn't seem the best choice for a getter.

phansys (changes requested):
I left some comments.
BTW, I think before merging, this PR should be splitted in 2 commits if you want to introduce the `convert_enum_to_string` option here.

src/AuditManager.php
michealmouner: how?
u can send it or not .. how will it break BC?

src/EventListener/CreateSchemaListener.php
michealmouner: `ENUM('XXX', 'YYY', 'ZZZZ')`

phansys (changes requested):

src/EventListener/CreateSchemaListener.php
phansys: IMO, then we could use `0 === strpos($sqlString, 'ENUM(')` in order to improve the search, since we already know that the string must start with the required substring.

michealmouner:
I created two separate PRs
one for ignoring specific property https://github.com/sonata-project/EntityAuditBundle/pull/401
one for converting enum to string https://github.com/sonata-project/EntityAuditBundle/pull/400

PR #398, opened 2021-05-31
Fix missing space in identifier WHERE clausule if the entity has multiple columns as the primary key

Vantomas:
## Subject

There is a problem if the entity has multiple columns as the primary key. In this case EntityAuditBundle ends with a SyntaxErrorException. 

`Doctrine\DBAL\Driver\Mysqli\Exception\ConnectionError: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '?AND itemid = ?' at line 1 in /Users/vantomas/Projekty/lms-dev/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/Mysqli/Exception/ConnectionError.php:19`

`Next Doctrine\DBAL\Exception\SyntaxErrorException: An exception occurred while executing 'UPDATE invoicecontents_audit SET docid = ? WHERE rev = ? AND docid = ?AND itemid = ?' with params [343765, 149657, 343765, 1]:`

I am targeting this branch, because it's without BC-break.

## Changelog

```markdown
### Fixed
- Fix missing space in identifier WHERE clausule if the entity has multiple columns as the primary key

```

### DIFF
--- a/src/EventListener/LogRevisionsListener.php
+++ b/src/EventListener/LogRevisionsListener.php
@@ -187,7 +187,7 @@ class LogRevisionsListener implements EventSubscriber
 
                     $params[] = $meta->reflFields[$idField]->getValue($entity);
 
-                    $sql .= 'AND '.$columnName.' = ?';
+                    $sql .= ' AND '.$columnName.' = ?';
                 }
 
                 $this->em->getConnection()->executeQuery($sql, $params, $types);


core23 (approved):

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #400, opened 2021-06-15
Convert enum to string option

michealmouner:
## Subject

I am targeting this branch, because I need to add a new features to the bundle
I added a boolean option to convert EnumToString incase you are using it in the DB (it causes problems)
```
```

example of created ENUM migration
```
$this->addSql('ALTER TABLE table_name ADD enum_column ENUM(\'option_1\', \'option_2\', \'option_3\') DEFAULT \'option_3\' NOT NULL');
$this->addSql('ALTER TABLE table_name ADD enum_column VARCHAR(255) DEFAULT \'option_3\'');
```

## Changelog

```markdown
### Added
two new options:

### Fixed
src/SimpleThings/EntityAudit/Utils/SimpleDiff.php:26 --> fix undefined variable

```

### DIFF
--- a/src/Utils/SimpleDiff.php
+++ b/src/Utils/SimpleDiff.php
@@ -50,6 +50,8 @@ class SimpleDiff
     {
         $ret = '';
         $diff = $this->diff(explode(' ', $old), explode(' ', $new));
+        $nmax = 0;
+        $omax = 0;
         foreach ($diff as $k) {
             if (\is_array($k)) {
                 $ret .= (!empty($k['d']) ? '<del>'.implode(' ', $k['d']).'</del> ' : '').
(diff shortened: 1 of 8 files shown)


VincentLanglet (dismissed):

VincentLanglet:
Can you take a look @phansys ?

src/Utils/SimpleDiff.php
```diff
@@ -50,6 +50,8 @@ public function htmlDiff($old, $new)
     {
         $ret = '';
         $diff = $this->diff(explode(' ', $old), explode(' ', $new));
+        $nmax = 0;
```
phansys, line 53: These variables seem to be unused in this method.

phansys (changes requested):
Could we add a test for this feature?

PR #410, opened 2021-07-20
DevKit updates for 1.x branch

SonataCI:


VincentLanglet (changes requested):
cc @phansys

VincentLanglet:
The solution we used last time we need something was to use `custom_bootstrap`
https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/blob/2.x/tests/custom_bootstrap.php

PR #413, opened 2021-07-24
Add QA checks through PHPStan and Psalm

phansys:


src/AuditReader.php
```diff
@@ -64,6 +64,9 @@ class AuditReader
      * Entity cache to prevent circular references.
      *
      * @var array
+     *
+     * @phpstan-template T of object
```
VincentLanglet, line 68: A template on a property seems weird, it's on the class or on a method generally.
All the `@phpstan-template T of object` might be moved to the class...

src/Collection/AuditedCollection.php
```diff
@@ -20,6 +20,12 @@
 use SimpleThings\EntityAudit\AuditReader;
 use SimpleThings\EntityAudit\Exception\AuditedCollectionException;
 
+/**
+ * @phpstan-template TKey
+ * @psalm-template TKey of array-key
+ * @psalm-template T
```
VincentLanglet, line 26: Why do you need different template for phpstan and psalm ?

src/ChangedEntity.php
```diff
@@ -17,6 +17,8 @@ class ChangedEntity
 {
     /**
      * @var string
+     *
+     * @phpstan-var class-string<T> $className
```
VincentLanglet, line 21: Tu use T, u have to defining the template on the class, and not only the construct i think

src/Collection/AuditedCollection.php
phansys: It was copied from `Collection`.

psalm-baseline.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<files psalm-version="4.11.1@e33492398bd4e5e2ab60e331d445979bd83feecd">
+    <file src="src/Controller/AuditController.php">
+        <UndefinedClass occurrences="1">
+            <code>Controller</code>
+        </UndefinedClass>
+    </file>
+    <file src="vendor/doctrine/orm/lib/Doctrine/ORM/Internal/SQLResultCasing.php">
```
VincentLanglet: this is weird

psalm-baseline.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<files psalm-version="4.11.1@e33492398bd4e5e2ab60e331d445979bd83feecd">
+    <file src="src/Controller/AuditController.php">
```
VincentLanglet, line 4: next major comment ?

phpstan-baseline.neon
```diff
@@ -0,0 +1,6 @@
+parameters:
+    ignoreErrors:
+        -
```
VincentLanglet: We try to add a comment to explain why we ignore an error.

Also, I'm curious why we need to ignore this, configuration was always fine for others projetcs.

psalm-baseline.xml
jordisala1991: The class is internal, but IMO this should be added on code not here (thats how we did on other cases)

VincentLanglet: But why do we have to ignore an error for a vendor file ?
`vendor/doctrine/orm/lib/Doctrine/ORM/Internal/SQLResultCasing.php`

jordisala1991: Oh, I don't know. I thought this was only because it was internal.

psalm-baseline.xml
phansys: Added.

psalm-baseline.xml
phansys: [`Doctrine\ORM\Internal\SQLResultCasing::getSQLResultCasing()`](https://github.com/doctrine/orm/blob/203cd6ecf256c43b4a9a7ca88a35ed7c87c585e1/lib/Doctrine/ORM/Internal/SQLResultCasing.php#L27) is using [`Doctrine\DBAL\Platforms\PostgreSQLPlatform`](https://github.com/doctrine/dbal/blob/75fe7631363e1957e087d03d09eb55b0bf8a493c/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php#L37-L44). This class was deprecated in "doctrine/dbal" 2.11.0 and removed in version 3.0.

phpstan-baseline.neon
phansys: I don't know why this configuration is working without errors in other projects.

psalm-baseline.xml
VincentLanglet: Yes, but psalm is not supposed to report errors in the vendor...

VincentLanglet: You need to ignoreFile part maybe https://github.com/sonata-project/SonataAdminBundle/blob/4.x/psalm.xml#L6-L9

phansys: Perfect. `vendor/` dir added to the ignore list :+1:

phpstan-baseline.neon
phansys: This issue was resolved after installing "phpstan/phpstan-symfony".
Thank you.

psalm.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<psalm xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="https://getpsalm.org/schema/config" errorLevel="6" resolveFromConfigFile="true" xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd" errorBaseline="psalm-baseline.xml">
+    <projectFiles>
+        <directory name="src"/>
```
VincentLanglet, line 4:
```suggestion
        <directory name="src"/>
        <directory name="tests"/>
```

composer.json
```diff
@@ -31,10 +31,17 @@
         "doctrine/doctrine-bundle": "^1.4 || ^2.2",
         "gedmo/doctrine-extensions": "^3.0",
         "matthiasnoback/symfony-dependency-injection-test": "^4.2.1",
+        "phpstan/extension-installer": "^1.1",
+        "phpstan/phpstan": "^0.12.93",
+        "phpstan/phpstan-symfony": "^0.12.44",
```
VincentLanglet, line 38: There is also a phpstan/phpstan-phpunit plugin

VincentLanglet: And maybe the phpstan/phpstan-doctrine should be added too.

composer.json
```diff
@@ -31,10 +31,17 @@
...
         "phpunit/phpunit": "^9.5",
         "symfony/cache": "^4.4 || ^5.3 || ^6.0",
         "symfony/http-foundation": "^4.4 || ^5.3 || ^6.0",
-        "symfony/phpunit-bridge": "^5.3 || ^6.0"
+        "symfony/http-kernel": "^4.4 || ^5.3 || ^6.0",
+        "symfony/phpunit-bridge": "^5.3 || ^6.0",
+        "symfony/security-core": "^4.4 || ^5.3 || ^6.0",
+        "symfony/var-dumper": "^4.4 || ^5.2",
+        "vimeo/psalm": "^4.8"
```
VincentLanglet: We should add the psalm-symfony and psalm-phpunit plugin (and psalm-doctrine)

VincentLanglet:
For the `Doctrine\Common\Collections\Collection::isEmpty is marked @psalm-immutable, but SimpleThings\EntityAudit\Collection\AuditedCollection::isEmpty is not marked @psalm-immutable (see https://psalm.dev/213` error, I know the reason. It's https://github.com/psalm/psalm-plugin-doctrine/issues/97

Doing something like https://github.com/psalm/psalm-plugin-doctrine/pull/99 should solve it.
Maybe a comment with the issue next to the `psalm` annotation could explain why this hacky solution was done.

phansys:
>Doing something like psalm/psalm-plugin-doctrine#99 should solve it.

As far as I can see, adding the `@psalm-mutation-free` annotation is required in almost all the methods in order to fix the reported problem.

BTW, do you know why the `@return` annotation is not respected in cases like this?
```
Cannot access array value on non-array variable $revisions of type iterable<mixed, mixed>
```

VincentLanglet:
> As far as I can see, adding the `@psalm-mutation-free` annotation is required in almost all the methods in order to fix the reported problem.

`@psalm-mutation-free` should only be needed for the isEmpty method.
 
> BTW, do you know why the `@return` annotation is not respected in cases like this?
> 
> ```
> Cannot access array value on non-array variable $revisions of type iterable<mixed, mixed>
> ```

Not really. Maybe adding a `assertIsArray` check solve the issue.

phansys:
>@psalm-mutation-free should only be needed for the isEmpty method.

I've added the annotations in the methods pointed by Psalm in different commits, in order to allow you to check what I mean.

>Not really. Maybe adding a assertIsArray check solve the issue.

Yes, it works adding the assertion, but I think the `@return` docblock should be enough for this case.

VincentLanglet:
> > @psalm-mutation-free should only be needed for the isEmpty method.
> 
> I've added the annotations in the methods pointed by Psalm in different commits, in order to allow you to check what I mean.

I see...

Seems like the solution of psalm-doctrine has some drawback.

I think the "best" way to handle this would be to
- Add `psalm-mutation-free` to `isEmpty`
- Also add `psalm-suppress` with the right error name to `isEmpty`

src/AuditManager.php
```diff
@@ -44,6 +44,9 @@ public function getConfiguration()
         return $this->config;
     }
 
+    /**
+     * NEXT_MAJOR: Use `\Doctrine\ORM\EntityManagerInterface` for argument 1.
+     */
```
franmomu:
```suggestion
    /**
     * NEXT_MAJOR: Use `\Doctrine\ORM\EntityManagerInterface` for argument 1.
     *
     * @return AuditReader
     */
```

I think with this there is no need of asserting array in tests (that usually happens because there is some missing return type declaration).

src/Collection/AuditedCollection.php
```diff
@@ -131,6 +139,9 @@ public function contains($element)
 
     /**
      * @return bool
+     *
+     * @psalm-mutation-free See https://github.com/psalm/psalm-plugin-doctrine/issues/97
```
franmomu, line 143: AFAIK this is not true here, we are mutating the internal state (`$initialized` property), do we need this psalm plugin?

src/AuditReader.php
```diff
@@ -227,20 +230,20 @@ public function find($className, $id, $revision, array $options = [])
             throw new NotAuditedException($className);
         }
 
-        /** @var ClassMetadataInfo|ClassMetadata $class */
-        $class = $this->em->getClassMetadata($className);
-        $tableName = $this->config->getTableName($class);
+        /** @var ClassMetadata<T> $classMetadata */
```
franmomu, line 233: I think this line is not needed

phansys: You're right. Removed.

src/AuditManager.php
phansys: The cases of the return types that weren't being respected, are those like this: https://github.com/sonata-project/EntityAuditBundle/blob/bf499c7a3ad12c1065159a384d1f53f5ae1e5133/src/AuditReader.php#L517.

src/Collection/AuditedCollection.php
phansys: See https://github.com/sonata-project/EntityAuditBundle/pull/413#discussion_r735852969.

VincentLanglet: @franmomu When using doctrine and psalm, I think natural to use the psalm-doctrine plugin.

I agree the @psalm-mutation-free is not valid. But this is the only known solution for the psalm-doctrine to keep the value of `isEmpty` in memory. This way when calling `first()` right after, the user doesn't need to do a `!== false` check. This was introduce here: https://github.com/psalm/psalm-plugin-doctrine/pull/94

src/AuditReader.php
phansys: I'm afraid I must restore these inline docblocks:
```------ -------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   src/AuditReader.php                                                                                                                                                 
 ------ -------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  335    Parameter #1 $className of method SimpleThings\EntityAudit\AuditReader<T of object>::createEntity() expects class-string<T of object>, class-string<object> given.  
  744    Parameter #1 $className of method SimpleThings\EntityAudit\AuditReader<T of object>::createEntity() expects class-string<T of object>, class-string<object> given.
```

tests/BaseTest.php
```diff
@@ -34,6 +35,8 @@ abstract class BaseTest extends TestCase
     protected static $conn;
 
     /**
+     * NEXT_MAJOR: Use `\Doctrine\ORM\EntityManagerInterface` instead.
```
jordisala1991, line 38: This is a test class, cant it be changes right away?

phansys: No, because the real classes are using `EntityManager` in its argument declarations.
By instance:
```
Parameter #1 $em of method SimpleThings\EntityAudit\AuditManager::createAuditReader() expects Doctrine\ORM\EntityManager, Doctrine\ORM\EntityManagerInterface given.
```

src/AuditManager.php
franmomu: yep, but before that call there is one to `$this->auditManager->createAuditReader`, if there is no return type it's interpreted as `mixed`, then `mixed->findX` will return `mixed` as well. I'll try later, but I guess that's the problem.

src/Collection/AuditedCollection.php
franmomu: > @franmomu When using doctrine and psalm, I think natural to use the psalm-doctrine plugin.
> 
> I agree the @psalm-mutation-free is not valid. But this is the only known solution for the psalm-doctrine to keep the value of `isEmpty` in memory. This way when calling `first()` right after, the user doesn't need to do a `!== false` check. This was introduce here: [psalm/psalm-plugin-doctrine#94](https://github.com/psalm/psalm-plugin-doctrine/pull/94)

I'm fine using it if it's solving or doing something, but I'd rather not use it without no reason (also taking into account that is forcing us to use wrong annotations and ignoring non-existing errors).

About `isEmpty` I have no idea how psalm works, but `@psalm-mutation-free` does not seem like a proper fix.

Anyway, just my opinion, not a blocker.

src/AuditManager.php
phansys: Yes. That's exactly the case. I've removed the assertions and added the return type docblock.
Thank you.

src/EventListener/CreateSchemaListener.php
```diff
@@ -45,9 +45,6 @@ public function __construct(AuditManager $auditManager)
         $this->metadataFactory = $auditManager->getMetadataFactory();
     }
 
-    /**
-     * @return array
-     */
```
franmomu: Removing this triggers a deprecation with Symfony 5.4: https://github.com/sonata-project/EntityAuditBundle/runs/4022339669?check_suite_focus=true#step:10:20

src/EventListener/LogRevisionsListener.php
```diff
@@ -92,9 +95,6 @@ public function __construct(AuditManager $auditManager)
         $this->metadataFactory = $auditManager->getMetadataFactory();
     }
 
-    /**
-     * @return array
-     */
```
franmomu: Removing this triggers a deprecation with Symfony 5.4: https://github.com/sonata-project/EntityAuditBundle/runs/4022339669?check_suite_focus=true#step:10:23

phansys:
@jordisala1991, @franmomu; could you please review again?

composer.json
```diff
@@ -31,10 +31,22 @@
...
+        "phpstan/phpstan-symfony": "^0.12.44",
         "phpunit/phpunit": "^9.5",
+        "psalm/plugin-phpunit": "^0.16.1",
+        "psalm/plugin-symfony": "^3.0",
         "symfony/cache": "^4.4 || ^5.3 || ^6.0",
         "symfony/http-foundation": "^4.4 || ^5.3 || ^6.0",
-        "symfony/phpunit-bridge": "^5.3 || ^6.0"
+        "symfony/http-kernel": "^4.4 || ^5.3 || ^6.0",
+        "symfony/phpunit-bridge": "^5.3 || ^6.0",
+        "symfony/security-core": "^4.4 || ^5.3 || ^6.0",
```
jordisala1991: this is already on require section

composer.json
```diff
@@ -31,10 +31,22 @@
...
+        "phpstan/phpstan-doctrine": "^0.12.44",
+        "phpstan/phpstan-phpunit": "^0.12.22",
+        "phpstan/phpstan-symfony": "^0.12.44",
         "phpunit/phpunit": "^9.5",
+        "psalm/plugin-phpunit": "^0.16.1",
+        "psalm/plugin-symfony": "^3.0",
         "symfony/cache": "^4.4 || ^5.3 || ^6.0",
         "symfony/http-foundation": "^4.4 || ^5.3 || ^6.0",
-        "symfony/phpunit-bridge": "^5.3 || ^6.0"
+        "symfony/http-kernel": "^4.4 || ^5.3 || ^6.0",
```
jordisala1991: this one too

composer.json
phansys: Good catch. Removed.

composer.json
```diff
@@ -31,10 +31,20 @@
...
+        "phpstan/phpstan-phpunit": "^0.12.22",
+        "phpstan/phpstan-symfony": "^0.12.44",
         "phpunit/phpunit": "^9.5",
+        "psalm/plugin-phpunit": "^0.16.1",
+        "psalm/plugin-symfony": "^3.0",
         "symfony/cache": "^4.4 || ^5.3 || ^6.0",
         "symfony/http-foundation": "^4.4 || ^5.3 || ^6.0",
-        "symfony/phpunit-bridge": "^5.3 || ^6.0"
+        "symfony/phpunit-bridge": "^5.3 || ^6.0",
+        "symfony/var-dumper": "^4.4 || ^5.2",
```
jordisala1991: you should use 4.4 5.3 and 6.0. To be consistent with the other symfony dependencies.

jordisala1991:
Please remember to open the dev-kit PR to add psalm and phpstan @phansys

PR #415, opened 2021-07-24
Make explicit some missing deps

phansys:
## Subject
Make explicit some missing deps.

I am targeting this branch, because these changes respect BC.

## Changelog

```markdown
### Fixed
- Missing dependencies required by this package.

```


VincentLanglet (approved):

composer.json
```diff
@@ -26,7 +30,7 @@
         "gedmo/doctrine-extensions": "^3.0",
         "matthiasnoback/symfony-dependency-injection-test": "^4.2.1",
         "symfony/cache": "^4.4 || ^5.2",
-        "symfony/framework-bundle": "^4.4 || ^5.2",
```
VincentLanglet, line 29: Why this removal ?

phansys: Class `Controller` was removed in `symfony/framework-bundle:5.x`:
https://github.com/sonata-project/EntityAuditBundle/blob/af61a7bfcec2052c381b5c3333a0ccbe6d7d7b3b/src/Controller/AuditController.php#L23-L37

VincentLanglet (approved):

phansys:
As stated in the checks for this PR, this package is not compatible with `symfony/symfony:"5.3.*"` and `symfony/symfony:"6.0.*"`.
IMO, we should try to provide forward compatible bridges or mark these builds as optional.
WDYT?

jordisala1991:
> As stated in the checks for this PR, this package is not compatible with `symfony/symfony:"5.3.*"` and `symfony/symfony:"6.0.*"`. IMO, we should try to provide forward compatible bridges or mark these builds as optional. WDYT?

I would remove symfony/framework as a dev dependency and instead add it as an extra build (with the variant thing). And ignore the errors of static analysis tools. the class is deprecated and is not suposed to be used or modified anymore, so not a big deal if it is ignored by the tools. wdyt?

composer.json
```diff
@@ -11,10 +11,14 @@
     "require": {
         "php": "^7.3 || ^8.0",
         "doctrine/collections": "^1.6",
+        "doctrine/common": "^3.1",
         "doctrine/dbal": "^2.11",
         "doctrine/event-manager": "^1.1",
         "doctrine/orm": "^2.9",
+        "doctrine/persistence": "^2.2",
         "symfony/config": "^4.4 || ^5.3 || ^6.0",
+        "symfony/http-kernel": "^4.4 || ^5.3 || ^6.0",
```
jordisala1991, line 21: why this as require dependency?

phansys:
>I would remove symfony/framework as a dev dependency and instead add it as an extra build (with the variant thing). And ignore the errors of static analysis tools. the class is deprecated and is not suposed to be used or modified anymore, so not a big deal if it is ignored by the tools. wdyt?

I agree.

composer.json
phansys: Because this package is a Symfony bundle: https://github.com/sonata-project/EntityAuditBundle/blob/5a442d2baeb566a45ecc2f244eb9a880e0b62000/src/SimpleThingsEntityAuditBundle.php#L14-L20

jordisala1991: Then dependency injection should be moved to, right?

.github/workflows/test.yaml
```diff
@@ -57,6 +57,10 @@ jobs:
                       allowed-to-fail: false
                       symfony-require: 6.0.*
                       variant: symfony/symfony:"6.0.*"
+                    - php-version: '8.0'
+                      dependencies: highest
+                      allowed-to-fail: false
+                      variant: symfony/framework-bundle
```
jordisala1991: imo we should make explicit the 4.4 version

phansys: Done.

VincentLanglet:
Thanks @phansys

PR #439, opened 2021-09-19
Add support for Symfony 6

jordisala1991:
## Subject

Part of: https://github.com/sonata-project/dev-kit/issues/1675
Dev-kit PR: https://github.com/sonata-project/dev-kit/pull/1684

## Changelog

```markdown
### Added
- Added explicit dependencies with Doctrine and Symfony.
- Added support for Symfony 6
### Removed
- Removed support for Symfony 5.2
```

### DIFF
--- a/src/Collection/AuditedCollection.php
+++ b/src/Collection/AuditedCollection.php
@@ -105,6 +105,9 @@ class AuditedCollection implements Collection
         $this->entities = new ArrayCollection();
     }
 
+    /**
+     * @return bool
+     */
     public function add($element)
     {
         throw new AuditedCollectionException('The AuditedCollection is read-only');
@@ -116,6 +119,9 @@ class AuditedCollection implements Collection
         $this->initialized = false;
     }
 
+    /**
+     * @return bool
+     */
     public function contains($element)
     {
         $this->forceLoad();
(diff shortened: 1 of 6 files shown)


phansys (commented):
About the "symfony/http-foundation" dependency, I think this should be optional, since this bundle can work without its controllers.

jordisala1991:
> About the "symfony/http-foundation" dependency, I think this should be optional, since this bundle can work without its controllers.

Yep, I added it to require dev for that matter.

phansys (approved):

src/Collection/AuditedCollection.php
```diff
@@ -105,6 +105,9 @@ public function __construct(AuditReader $auditReader, $class, ClassMetadataInfo
         $this->entities = new ArrayCollection();
     }
 
+    /**
+     * @return bool
```
VincentLanglet, line 109: Why can't we rely on interface phpdoc ?

VincentLanglet (approved):

src/Collection/AuditedCollection.php
jordisala1991: If you don't add it here, you get deprecation from symfony. This way we silence it (and we know that at some point we will need to add it to the method directly without phpdoc.

PR #444, opened 2021-10-03
Allow Doctrine DBAL 3

jordisala1991:
## Subject

## Changelog

```markdown
### Added
- Added support for Doctrine DBAL 3.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         "php": "^7.3 || ^8.0",
         "doctrine/collections": "^1.6",
         "doctrine/common": "^3.0",
-        "doctrine/dbal": "^2.13",
+        "doctrine/dbal": "^2.13 || ^3.0",
         "doctrine/event-manager": "^1.1",
         "doctrine/orm": "^2.10",
         "doctrine/persistence": "^2.0",
--- a/src/AuditReader.php
+++ b/src/AuditReader.php
@@ -458,7 +458,7 @@ class AuditReader
             }
 
             $query = 'SELECT '.$columnList.' FROM '.$tableName.' e '.$joinSql.' WHERE '.$whereSQL;
-            $revisionsData = $this->em->getConnection()->executeQuery($query, $params);
+            $revisionsData = $this->em->getConnection()->fetchAllAssociative($query, $params);
 
             foreach ($revisionsData as $row) {
                 $id = [];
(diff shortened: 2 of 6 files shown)


jordisala1991:
Not sure with any of this changes, just tried to follow deprecation messages. This bundle blocks the doctrine orm admin bundle. @phansys or @VincentLanglet do you know why test fail?

simonberger:
@jordisala1991 The changes are looking good to me. Didn't check if they are complete.

To the tests failures it is probably this:
https://github.com/doctrine/dbal/blob/2.13.x/lib/Doctrine/DBAL/Schema/Column.php#L78-L89

Which means already a problem in the current releases and another deprecation in dbal 2.X which has to be solved.

simonberger:
Just a side note:
I would prefer tests to fail when hitting a deprecation or at least displaying it. I saw no such output in the previous runs of PHPUnit.

VincentLanglet:
https://github.com/sonata-project/EntityAuditBundle/pull/446 might help @jordisala1991

simonberger (changes requested):
The tests are failing because of:
```php
$revisionsData = $this->em->getConnection()->executeQuery($query, $params);
```
https://github.com/sonata-project/EntityAuditBundle/pull/444/files#diff-d3e6d43afbe921b0b71e48fdc65d0402668ac18f5306f927ef57f81d3cb1d7abL459

It works in 2.13 and 3.1 like this:

```php
$revisionsData = $this->em->getConnection()->fetchAllAssociative($query, $params);
```



Alternatively do `executeQuery` with single `fetchAssociative` in a while loop.

I am still irritated why the existing code did not fail in 2.13 and when it has been changed. Some forwardCompatibility classes are executed there, so probably it was a change many versions ago which is not well documented.

jordisala1991:
I think it would be nice to get your review too @simonberger

phansys (approved):

simonberger (approved):
Changes in this PR look good to me as well.
Approved with minor comment:

The use of the `SQLResultCasing` trait is not optimal. Changes or a removal could happen in every version.
Considering the minimal code it provides the trait probably should be copied into this library. Didn't mention it in the other PR because I thought the solution here will be different.

jordisala1991:
> Changes in this PR look good to me as well.
> 
> Approved with minor comment:
> 
> 
> 
> The use of the `SQLResultCasing` trait is not optimal. Changes or a removal could happen in every version.
> 
> Considering the minimal code it provides the trait probably should be copied into this library. Didn't mention it in the other PR because I thought the solution here will be different.

Do you want to provide a PR with that changes? I think we can add that trait to our library too

PR #446, opened 2021-10-16
Fix dbal deprecation of passing unknown options to a `Column`

simonberger:
After allowing Dbal 3 in https://github.com/sonata-project/EntityAuditBundle/pull/444 several tests failed because unknown option had been passed to a `Column`. This was deprecated at some point.

## Changelog

```markdown
### Fixed
- Avoid passing unknown options to a column during its creation.
```

### DIFF
--- a/src/EventListener/CreateSchemaListener.php
+++ b/src/EventListener/CreateSchemaListener.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace SimpleThings\EntityAudit\EventListener;
 
 use Doctrine\Common\EventSubscriber;
+use Doctrine\DBAL\Schema\Column;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Types\Types;
@@ -83,10 +84,7 @@ class CreateSchemaListener implements EventSubscriber
         );
 
         foreach ($entityTable->getColumns() as $column) {
-            $revisionTable->addColumn($column->getName(), $column->getType()->getName(), array_merge(
-                $column->toArray(),
-                ['notnull' => false, 'autoincrement' => false]
-            ));
+            $this->addColumnToTable($column, $revisionTable);
         }
         $revisionTable->addColumn($this->config->getRevisionFieldName(), $this->config->getRevisionIdFieldType());
         $revisionTable->addColumn($this->config->getRevisionTypeFieldName(), Types::STRING, ['length' => 4]);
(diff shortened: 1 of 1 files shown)


src/EventListener/CreateSchemaListener.php
```diff
@@ -83,10 +84,13 @@ public function postGenerateSchemaTable(GenerateSchemaTableEventArgs $eventArgs)
...
-                $column->toArray(),
-                ['notnull' => false, 'autoincrement' => false]
-            ));
+            $columnOptions = array_filter($column->toArray(), static function($columnOption, $columnKey) {
+                return method_exists(Column::class, 'set'.$columnKey);
+            }, ARRAY_FILTER_USE_BOTH);
+            $columnOptions['notnull'] = false;
+            $columnOptions['autoincrement'] = false;
+
+            $revisionTable->addColumn($column->getName(), $column->getType()->getName(), $columnOptions);
```
simonberger: This does not pass `customSchemaOptions` or `platformOptions` (also ignored in dbal 2.X) but should it be passed?
Could be changed with:
```php
$columnOptions['platformOptions'] = $column->getPlatformOptions();
$columnOptions['customSchemaOptions'] = $column->setCustomSchemaOptions();
```

simonberger:
I am not very happy with this fix. Imo the toArray method should not be used or should pass the platformOptions and customSchemaOptions in a sub array instead of merging them on the root level.

Another solution could be enabling the `Table` in dbal to add a column from another column to not need so much class knowledge in other libraries.

The parameter name options confused me most, aren't it somewhat just the properties of the column-class?

src/EventListener/CreateSchemaListener.php
```diff
@@ -83,10 +84,13 @@ public function postGenerateSchemaTable(GenerateSchemaTableEventArgs $eventArgs)
         );
 
         foreach ($entityTable->getColumns() as $column) {
-            $revisionTable->addColumn($column->getName(), $column->getType()->getName(), array_merge(
-                $column->toArray(),
-                ['notnull' => false, 'autoincrement' => false]
-            ));
+            $columnOptions = array_filter($column->toArray(), static function ($columnOption, $columnKey) {
+                return method_exists(Column::class, 'set'.$columnKey);
```
jordisala1991: why don't we just take the 'name' key option?

Or we could add those options by hand looking at the toArray method there are not much things to be done (it would also save us that array_merge that we do not need:

```
        return array_merge([
            'name'          => $this->_name,
            'type'          => $this->_type,
            'default'       => $this->_default,
            'notnull'       => $this->_notnull,
            'length'        => $this->_length,
            'precision'     => $this->_precision,
            'scale'         => $this->_scale,
            'fixed'         => $this->_fixed,
            'unsigned'      => $this->_unsigned,
            'autoincrement' => $this->_autoincrement,
            'columnDefinition' => $this->_columnDefinition,
            'comment' => $this->_comment,
        ], $this->_platformOptions, $this->_customSchemaOptions);
```

simonberger: > why don't we just take the 'name' key option?

The `name` result from `toArray()` can have a different value because of the namespace.

> Or we could add those options by hand looking at the toArray method there are not much things to be done (it would also save us that array_merge that we do not need

That is possible but it would mean the knowledge of all available options now and in the future just to move this column. That's some kind of business logic I would not want to move into my library. 
If this is fine for you at this place I can easily implement it.

jordisala1991: wdyt @phansys ? 

Im not a big fan of checking for method exist. But building the array here means we know all the options (Not sure if that's bad or not, if the options are correctly shaped on the phpdoc is not a big deal tbh). The ideal world would be having a method that let us call addColumn giving another column, but IMO we can't wait for that to happen on dbal, we should fix this and maybe if that happens improve later.

simonberger: Copied option by option now instead. Copying customSchemaOptions and platformOptions as well. I doubt those had been left out intentionally.

jordisala1991:
To me this solution looks great, can you fix php cs fixer build?

jordisala1991 (approved):

phansys (approved):

phansys (commented):
IMO, this PR is RTM.
Could you please use the PR template in order to provide the required info?

jordisala1991:
> IMO, this PR is RTM. Could you please use the PR template in order to provide the required info?

Fixed

jordisala1991:
Thanks @simonberger

PR #451, opened 2021-10-20
Avoid some deprecations

jordisala1991:
## Subject

I am targeting this branch, because this is pedantic.

Will help with the upgrade to 3.0 of dbal

## Changelog

```markdown
### Removed
- Removed support for Doctrine ORM < 2.10
- Removed support for Doctrine DBAL < 2.13
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,9 +11,11 @@
     "require": {
         "php": "^7.3 || ^8.0",
         "doctrine/collections": "^1.6",
-        "doctrine/dbal": "^2.11",
+        "doctrine/common": "^3.0",
+        "doctrine/dbal": "^2.13",
         "doctrine/event-manager": "^1.1",
-        "doctrine/orm": "^2.9",
+        "doctrine/orm": "^2.10",
+        "doctrine/persistence": "^2.0",
         "symfony/config": "^4.4 || ^5.3 || ^6.0",
         "twig/twig": "^2.6 || ^3.0"
     },
@@ -23,7 +25,6 @@
         "symfony/framework-bundle": "<4.4"
     },
     "require-dev": {
-        "doctrine/cache": "^1.11 || ^2.0",
         "doctrine/doctrine-bundle": "^1.4 || ^2.2",
         "gedmo/doctrine-extensions": "^3.0",
         "matthiasnoback/symfony-dependency-injection-test": "^4.2.1",
--- a/src/AuditReader.php
+++ b/src/AuditReader.php
@@ -18,13 +18,13 @@ use Doctrine\DBAL\Exception;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Internal\SQLResultCasing;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\Mapping\QuoteStrategy;
 use Doctrine\ORM\ORMException;
 use Doctrine\ORM\PersistentCollection;
 use Doctrine\ORM\Persisters\Entity\EntityPersister;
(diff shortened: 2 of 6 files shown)


src/AuditReader.php
```diff
@@ -18,6 +18,7 @@
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Internal\SQLResultCasing;
```
jordisala1991, line 21: This class is internal to ORM, but I think we might be using it for now, since it is the only non deprecated way I found to have this function

composer.json
```diff
@@ -11,9 +11,11 @@
     "require": {
         "php": "^7.3 || ^8.0",
         "doctrine/collections": "^1.6",
-        "doctrine/dbal": "^2.11",
+        "doctrine/common": "^3.0",
+        "doctrine/dbal": "^2.13",
         "doctrine/event-manager": "^1.1",
-        "doctrine/orm": "^2.9",
+        "doctrine/orm": "^2.10",
```
jordisala1991, line 17: Not a big fan of bumping this requirement to 2.10 but this is the version where the SqlResultCasing was introduced. wdyt @sonata-project/contributors ?

VincentLanglet: Bumping a minor is fine to me. People should be able to update without any issues.

VincentLanglet (approved):

composer.json
VincentLanglet: But I would recommend to consider this as a minor and not a pedantic

jordisala1991: Changed to minor and added changelog

jordisala1991:
Please review @sonata-project/contributors

src/AuditReader.php
```diff
@@ -606,7 +608,7 @@ public function getCurrentRevision($className, $id)
         $query = 'SELECT e.'.$this->config->getRevisionFieldName().' FROM '.$tableName.' e '.
                         ' WHERE '.$whereSQL.' ORDER BY e.'.$this->config->getRevisionFieldName().' DESC';
 
-        return $this->em->getConnection()->fetchColumn($query, array_values($id));
+        return $this->em->getConnection()->fetchOne($query, array_values($id));
```
phansys, line 611: Regardless the description at [`Connection::fetchColumn()`](https://github.com/doctrine/dbal/blob/85f40b1773ff4b0a922916cc83d10a52325f0d71/lib/Doctrine/DBAL/Connection.php#L641-L665) and [`Connection::fetchOne()`](https://github.com/doctrine/dbal/blob/85f40b1773ff4b0a922916cc83d10a52325f0d71/lib/Doctrine/DBAL/Connection.php#L717-L740), could you please confirm this replacement is not modifying the returned value?
`Connection::fetchOne()` seems to be intended to return an entire row instead of one column.

jordisala1991: Yes, indeed this change is wrong, do you know how should be written instead?

phansys: No, but let me do some checks and I'll try to find an answer.

simonberger: The change is fine and it is execute multiple times during the tests.
It does not return a row.

simonberger (approved):

src/AuditReader.php
jordisala1991: I will debug the value returned on test with both the old and the new method. The api seems a bit confusing on dbal side tho.

phansys: I've checked the behavior, forget about my concern.
I can confirm that `Connection::fetchOne()` is returning only the first selected column in the resulting row.
* Using `SELECT o.id, o.someColumn FROM ...`, the result is the value in `o.id`.
* Using `SELECT o.someColumn, o.id FROM ...`, the result is the value in `o.someColumn`.

phansys (approved):

phansys:
Thank you @jordisala1991!

Issue #370, opened 2021-03-10
Audit tables are being generated for all entities, not just the ones specified in config


alexandrubau:
The command `php bin/console d:s:u --dump-sql` dumps an SQL that would generate audit tables for all entities in my project even though I specified only one entity in config.

```yaml
simple_things_entity_audit:
  audited_entities:
    - App\Entity\Page
``` 

In [docs](https://sonata-project.org/bundles/doctrine-orm-admin/3-x/doc/reference/audit.html) it says that a `force.audit` option would disable audit for all entities but that option does not exist. 

Using `sonata-project/EntityAuditBundle` v1.1.0

phansys:
I think this issue is related to #199,

alexandrubau:
Thank you very much @phansys ! I was specifying `force.audit` option for `simple_things_entity_audit` instead of `sonata_doctrine_orm_admin`

bobemoe:
Just stumbled across this too, the fix works but its seems a but unintuitive default behaviour, I inferred from the README here that 'Load extension "simple_things_entity_audit" and specify the audited entities' was required and would take affect straight away.

Maybe a mention of this in the README would be good? 

NOTE: This (is not required? and) will not take affect until unless `force.audit` option for `sonata_doctrine_orm_admin` is set to `false`.

Issue #199, opened 2016-08-01
audited_entities configured but not used


Jerome1337:
I'm using this bundle with the Symfony 2.8 app and Doctrine ORM 2.5
I have install, enable and configure this bundle with the entities I want to audit.

```
# SimpleThings Entity Audit
simple_things_entity_audit:
    audited_entities:
        - AppBundle\Entity\User
        - AppBundle\Entity\Server
        - AppBundle\Entity\IpV4
        - AppBundle\Entity\IpV6
        - AppBundle\Entity\Purchase
```

The bundle should now only audit this entities and create only 6 tables (for this 5 entities and the revisions table).

But when I use `./app/console doctrine:schema:update --force` this create `_audit` tables for every single table of my database.

Did I forget something or this is a common bug ?

bendavies:
@Jerome1337 I can't replicate this. There is code in the schema listener to prevent against this:
https://github.com/simplethings/EntityAudit/blob/2d8de80e851cdac1071f54fd710bad4b7ae84f51/src/SimpleThings/EntityAudit/EventListener/CreateSchemaListener.php#L60-L76

Can you please reopen this if you can see what the issue is, but i'm going to close for now.

ceesvanegmond:
@Jerome1337 Do you solved the problem. I'm having the exact same problem @bendavies. Currently investigating...

ceesvanegmond:
@bendavies Found the problem, problem occurs when using Sonata Admin. Do you have this code enabled? This should solve it.

```yml
sonata_doctrine_orm_admin:
    audit:
        force: false
```

Jerome1337:
@ceesvanegmond Exactly this solve the creation problem.

PR #455, opened 2021-10-29
update docs to show auditied_entites needs enabling

bobemoe:
## Subject

Try and clear up the confusion described in #199 and #370

I am targeting this branch, because no BC break, just docs.


README.md
```diff
@@ -82,6 +82,8 @@ return [
 
 Load extension "simple_things_entity_audit" and specify the audited entities
 
+NOTE: All Entities are audited by default, even specifying here will not take affect unless `force.audit` option for `sonata_doctrine_orm_admin` is set to `false`. See https://github.com/sonata-project/EntityAuditBundle/issues/199
```
phansys, line 85: If this behavior is caused by "sonata-project/doctrine-orm-admin-bundle", I think this note should be added in its docs instead.

VincentLanglet:
Closing since it should be inside doctrineORMAdmin instead

PR #458, opened 2021-11-28
DevKit updates for 1.x branch

SonataCI:


src/Collection/AuditedCollection.php
```diff
@@ -397,6 +400,7 @@ public function offsetUnset($offset): void
     /**
      * @return int
      */
+    #[\ReturnTypeWillChange]
```
jordisala1991, line 403: We will have to add this to some method to avoid a self deprecation on php 8.1

VincentLanglet (dismissed):

PR #468, opened 2022-01-03
DevKit updates for 1.x branch

SonataCI:


jordisala1991:
Maybe @phansys can review this changes, I do not know enough of this bundle to be sure if the changes make sense or not.

VincentLanglet:
> Maybe @phansys can review this changes, I do not know enough of this bundle to be sure if the changes make sense or not.

It's not needed to understand how this library works. Personnaly I don't.

Priori to this change, the `$entities` property was an array of array or object.
And when needed the object were loaded.

But phpstan/psalm were complaining because it was requiring callable which support both array and object even if the callable were only use after object were loaded, so the callable was only used on object.

Instead of having both array and object in the same property, I used two different now.

is_array($entities[$key]) become isset($loadedEntities[$key]).

jordisala1991:
> > Maybe @phansys can review this changes, I do not know enough of this bundle to be sure if the changes make sense or not.
> 
> 
> 
> It's not needed to understand how this library works. Personnaly I don't.
> 
> 
> 
> Priori to this change, the `$entities` property was an array of array or object.
> 
> And when needed the object were loaded.
> 
> 
> 
> But phpstan/psalm were complaining because it was requiring callable which support both array and object even if the callable were only use after object were loaded, so the callable was only used on object.
> 
> 
> 
> Instead of having both array and object in the same property, I used two different now.
> 
> 
> 
> is_array($entities[$key]) become isset($loadedEntities[$key]).

Well, without this explanation it wasnt that easy to understand 😅.

Let me check the Pr again.

src/Collection/AuditedCollection.php
```diff
@@ -75,12 +75,24 @@ class AuditedCollection implements Collection
...
-     * @phpstan-var Collection<TKey, T|array{keys: array, rev: string|int}>
+     * @phpstan-var Collection<TKey, array{keys: array, rev: string|int}>
      */
     protected $entities;
 
+    /**
+     * Entity collection. If can be:
+     * - empty, if the collection has not been initialized yet
+     * - store entity
+     * - contain audited entity.
```
jordisala1991: Maybe this comment is not right anymore? Both properties has the same , but you kinda splitted how the previous property works, right?

VincentLanglet: I tried to improve the comment

jordisala1991 (approved):
LGTM

PR #469, opened 2022-01-05
Fixed join columns to be converted to the correct type

webmozart:
## Subject

Currently, `AuditReader` fails when handling `ToOne` associations with Symfony Uuids as primary keys. This is fixed by converting the join column value with the join column's type as is already done in line 834 for regular fields.

I am targeting this branch, because this is a simple patch.

## Changelog

```markdown
### Fixed
- Fixed `AuditReader` to process to-many associations using IDs with custom types 
```


jordisala1991:
The failure in psalm might get solved in #468

VincentLanglet (approved):

jordisala1991 (approved):

PR #473, opened 2022-01-16
Split issue tests

franmomu:
We had all test related to issues in one big test, this split them in individual tests so it's easier to add a new one. ~Probably the same should be done with [RelationTest](https://github.com/sonata-project/EntityAuditBundle/blob/0d27594275bc3255b788f34bca54d4be2a53ac7a/tests/RelationTest.php).~

I've also added some phpdoc types.


tests/Fixtures/Issue/Issue156Contact.php
```diff
@@ -69,7 +68,7 @@ public function removeTelephoneNumber(Issue156ContactTelephoneNumber $telephoneN
     }
 
     /**
-     * @return ArrayCollection|Issue156ContactTelephoneNumber[]
+     * @return iterable<int, Issue156ContactTelephoneNumber>
```
VincentLanglet, line 71: Dunno why iterable is used as typehint, it could be `Collection` instead isn't it ?

VincentLanglet (approved):

jordisala1991 (approved):

PR #474, opened 2022-01-26
1.7.0

VincentLanglet:


VincentLanglet:
@orklah any idea why I'm getting
```
 MethodSignatureMustProvideReturnType - src/Collection/AuditedCollection.php:127:21 - Method SimpleThings\EntityAudit\Collection\AuditedCollection::add must have a return type signature! (see https://psalm.dev/282)
    public function add($element)
```
errors ?

Is there a regression in psalm 4.19.0 ?

Edit: sticking the version to 4.18 solve the issue, so yes it's a regression.

Edit: seems like the doctrine plugin stubs are considered as native files.

jordisala1991 (approved):

Issue #143, opened 2015-08-13
Warning: array_flip(): Can only flip STRING and INTEGER values!


maxpowel:
Hello!

I just installed and configured the bundle but when clearing cache I got this warning:

Warning: array_flip(): Can only flip STRING and INTEGER values! in /Users/maxpowel/erp/vendor/simplethings/entity-audit-bundle/src/SimpleThings/EntityAudit/Metadata/MetadataFactory.php line 32 

My symfony version is:
Symfony version 2.5.12 - app/dev/debug

tolry:
ping @DavidBadura @andrewtch

DavidBadura:
Sorry for the late response. Can you send us your configuration? We can not reproduce this problem.

DavidBadura:
We can not reproduce it. If the problem still exists (in the latest version), please create a new ticket with your configuration.

SanderVerkuil:
It appears that the [Doctrine ORM Bundle](https://github.com/sonata-project/SonataDoctrineORMAdminBundle) iterates over all the registered entities and if they do not have a  'audit = false' in the admin definition, add it to the registered audit entities. This can be turned of by:
```yaml
# app/config/config.yml
sonata_doctrine_orm_admin:
    audit:
        force: false
```

While this might be an error on behalf of the Doctrine ORM bundle, I think that it'd be wise to filter the locations anyhow. A fix has been proposed in #285.

PR #486, opened 2022-04-23
Psalm 4

franmomu:
## Subject

Raises Psalm to level 4 fixing some types on the way

I am targeting this branch, because fixes some phpdoc types.

## Changelog

```markdown
### Fixed
- Fixed some phpdoc types
```

### DIFF
--- a/src/AuditReader.php
+++ b/src/AuditReader.php
@@ -563,6 +563,7 @@ class AuditReader
     }
 
     /**
+     * NEXT_MAJOR: Add NoRevisionFoundException as possible exception.
      * Gets the current revision of the entity with given ID.
      *
      * @param string           $className
@@ -571,7 +572,7 @@ class AuditReader
      * @throws NotAuditedException
      * @throws Exception
      *
-     * @return int|string
+     * @return int|string|null
      *
      * @phpstan-param class-string<T> $className
      */
@@ -606,7 +607,15 @@ class AuditReader
         $query = 'SELECT e.'.$this->config->getRevisionFieldName().' FROM '.$tableName.' e '.
                         ' WHERE '.$whereSQL.' ORDER BY e.'.$this->config->getRevisionFieldName().' DESC';
 
-        return $this->em->getConnection()->fetchOne($query, array_values($id));
+        $revision = $this->em->getConnection()->fetchOne($query, array_values($id));
+
+        if (false === $revision) {
+            // NEXT_MAJOR: Remove next line and uncomment the following one, also remove "null" as possible return type.
+            return null;
+            // throw new NoRevisionFoundException($className, $id, null);
+        }
+
+        return $revision;
     }
 
     /**
--- a/src/Collection/AuditedCollection.php
+++ b/src/Collection/AuditedCollection.php
@@ -122,7 +122,7 @@ class AuditedCollection implements Collection
     }
 
     /**
-     * @return bool
+     * @return true
      */
     public function add($element)
     {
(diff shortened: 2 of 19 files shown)


src/EventListener/LogRevisionsListener.php
```diff
@@ -308,7 +308,7 @@ private function getOriginalEntityData(object $entity): array
     }
 
     /**
-     * @return string|int
+     * @return string|int|false
```
VincentLanglet: if you're returning false, it means you're setting false to the property too.

Also `false` doesn't seems to be a supported return type because it's later use as a param in the query.

Maybe we should throw an exception if the value is false ?

src/AuditReader.php
```diff
@@ -571,7 +571,7 @@ public function findRevisions($className, $id)
      * @throws NotAuditedException
      * @throws Exception
      *
-     * @return int|string
+     * @return int|string|false
```
VincentLanglet: should we prefer null ?
or throw an exception in next major ?

src/Collection/AuditedCollection.php
```diff
@@ -122,7 +122,7 @@ public function __construct(AuditReader $auditReader, $class, ClassMetadataInfo
     }
 
     /**
-     * @return bool
+     * @return true
```
VincentLanglet, line 125: why this change ? this method return nothing.

src/Metadata/MetadataFactory.php
```diff
@@ -18,12 +18,12 @@ class MetadataFactory
     /**
      * @var string[]
      *
-     * @phpstan-var class-string[]
+     * @phpstan-var array<class-string, int|string>
```
VincentLanglet: array<class-string, int|string> is not fully compatible with string[], so maybe the @var need to be updated too.

src/Metadata/MetadataFactory.php
```diff
@@ -18,12 +18,12 @@ class MetadataFactory
...
      */
     private $auditedEntities = [];
 
     /**
-     * @phpstan-param array<class-string, mixed> $auditedEntities
+     * @phpstan-param class-string[] $auditedEntities
```
VincentLanglet, line 26: there is something weird according to phpstan

src/Metadata/MetadataFactory.php
```diff
@@ -16,36 +16,36 @@
...
      */
     public function __construct(array $auditedEntities)
     {
-        $this->auditedEntities = array_flip(array_filter($auditedEntities, static function ($record): bool {
-            return \is_string($record) || \is_int($record);
-        }));
+        $this->auditedEntities = array_flip($auditedEntities);
```
franmomu: Apparently this was done in https://github.com/sonata-project/EntityAuditBundle/issues/143 because of SonataDoctrineORMAdminBundle, but [looking at the tests there](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1652), looks like there are no `null` values. ~We could still add an `array_filter` call just in case~, but if there are `null` values, we should fix it in the doctrine orm admin bundle.

src/Collection/AuditedCollection.php
franmomu: To have the same return type as the interface: https://github.com/doctrine/collections/blob/c1ec9823f9250fefb274cccf62970b5c167e6bc7/lib/Doctrine/Common/Collections/Collection.php#L34-L42

src/AuditReader.php
franmomu: I'll do that

VincentLanglet (approved):

PR #487, opened 2022-04-23
PHPStan level 6

franmomu:
## Subject

Adds more specific phpdoc types, I've marked as `minor` because it was more adding missing types, but if you think `patch` is fine, it can be changed.

There are two commits because I'm not sure about the second one. It's necessary because `AuditReader` is generic, but I'm not sure if the whole class should be generic or only some of the method 🤔 

I am targeting this branch, because these changes are BC.

## Changelog

```markdown
### Fixed
- Specify iterable types
```

PR #488, opened 2022-04-23
PHPStan level 6 with `AuditReader` not generic

franmomu:
## Subject

Alternative to https://github.com/sonata-project/EntityAuditBundle/pull/487 making `AuditReader` not generic.

Raising PHPStan level after this to level 7 shows 23 errors.

## Changelog

```markdown
### Added
- Specify iterable types
### Changed
- Make `AuditReader` not generic
```

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,5 +1,5 @@
 parameters:
-    level: 5
+    level: 6
     paths:
         - src
         - tests
@@ -12,3 +12,5 @@ parameters:
     checkMissingIterableValueType: true
     checkMissingVarTagTypehint: true
     checkMissingTypehints: true
+    ignoreErrors:
+        - '#^Method SimpleThings\\EntityAudit\\Tests\\[a-zA-Z0-9_\\]+Test::(data|provide)[a-zA-Z0-9_]+\(\) return type has no value type specified in iterable type [a-zA-Z0-9_]+\.$#'
--- a/src/AuditReader.php
+++ b/src/AuditReader.php
@@ -206,9 +206,10 @@ class AuditReader
      * returns last revision INCLUDING "DEL" revision. If you want to throw exception instead, set
      * $threatDeletionAsException to true.
      *
-     * @param string           $className
-     * @param int|string|array $id
-     * @param int|string       $revision
+     * @param string                                    $className
+     * @param int|string|array<string, int|string>      $id
+     * @param int|string                                $revision
+     * @param array{threatDeletionsAsExceptions?: bool} $options
      *
      * @throws DeletedException
      * @throws NoRevisionFoundException
@@ -377,7 +378,7 @@ class AuditReader
      *
      * @param string|int $revision
      *
-     * @return ChangedEntity[]
+     * @return ChangedEntity<object>[]
      */
     public function findEntitesChangedAtRevision($revision)
     {
@@ -396,7 +397,7 @@ class AuditReader
      * @throws ORMException
      * @throws \RuntimeException
      *
-     * @return ChangedEntity[]
+     * @return ChangedEntity<object>[]
      */
     public function findEntitiesChangedAtRevision($revision)
     {
@@ -508,8 +509,8 @@ class AuditReader
     /**
      * Find all revisions that were made of entity class with given id.
      *
-     * @param string           $className
-     * @param int|string|array $id
+     * @param string                               $className
+     * @param int|string|array<string, int|string> $id
      *
      * @throws NotAuditedException
      * @throws Exception
@@ -566,8 +567,8 @@ class AuditReader
      * NEXT_MAJOR: Add NoRevisionFoundException as possible exception.
      * Gets the current revision of the entity with given ID.
      *
-     * @param string           $className
-     * @param int|string|array $id
+     * @param string                               $className
+     * @param int|string|array<string, int|string> $id
      *
      * @throws NotAuditedException
      * @throws Exception
@@ -677,8 +678,8 @@ class AuditReader
     }
 
     /**
-     * @param string           $className
-     * @param int|string|array $id
+     * @param string                               $className
+     * @param int|string|array<string, int|string> $id
      *
      * @throws DeletedException
      * @throws NoRevisionFoundException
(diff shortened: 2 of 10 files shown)


phpstan.neon.dist
```diff
@@ -12,3 +12,5 @@ parameters:
     checkMissingIterableValueType: true
     checkMissingVarTagTypehint: true
     checkMissingTypehints: true
+    ignoreErrors:
+        - '#Test::(data|provide)[a-zA-Z0-9_]+\(\) return type has no value type specified in iterable type#'
```
phansys: Just a less greedy suggestion:
```suggestion
        - '#^Method SimpleThings\\EntityAudit\\Tests\\[a-zA-Z0-9_\\]+Test::(data|provide)[a-zA-Z0-9_]+\(\) return type has no value type specified in iterable type [a-zA-Z0-9_]+\.$#'
```

src/AuditReader.php
```diff
@@ -788,6 +791,8 @@ protected function getEntityPersister($className)
      *
      * @phpstan-param class-string<T> $className
      * @phpstan-return T|null
+     *
+     * @template T of object
```
phansys: IMO, we should declare the template before its uses.

VincentLanglet:
Removing the template would create issues for the sonata compatibility since there is one in the SonataAdmin interface
https://github.com/sonata-project/SonataAdminBundle/blob/4.x/src/Model/AuditReaderInterface.php#L19

I wonder if we shouldn't have both
```
/**
 * @phpstan-template T of object
 */
class AuditReader
{
    /**
     * @phpstan-template TT of T
     * @phpstan-param class-string<TT>
     * @phpstan-return TT
     */
    public function find($classname): object
    {
    }
}
```

This way we can define an AuditReader<Foo|Bar>, but then
if we use find(Foo::class), it will be understood as a Foo object and not only Foo|Bar.

WDYT ?

We should update https://github.com/sonata-project/SonataAdminBundle/blob/4.x/src/Model/AuditReaderInterface.php phpdoc then and the same could be done for the ModelManager https://github.com/sonata-project/SonataAdminBundle/blob/4.x/src/Model/ModelManagerInterface.php#L55

franmomu:
> Removing the template would create issues for the sonata compatibility since there is one in the SonataAdmin interface https://github.com/sonata-project/SonataAdminBundle/blob/4.x/src/Model/AuditReaderInterface.php#L19
> 
> I wonder if we shouldn't have both
> 
> ```
> /**
>  * @phpstan-template T of object
>  */
> class AuditReader
> {
>     /**
>      * @phpstan-template TT of T
>      * @phpstan-param class-string<TT>
>      * @phpstan-return TT
>      */
>     public function find($classname): object
>     {
>     }
> }
> ```
> 
> This way we can define an AuditReader<Foo|Bar>, but then if we use find(Foo::class), it will be understood as a Foo object and not only Foo|Bar.
> 
> WDYT ?

IMHO that seems overcomplicated.

I understand that `AuditReader::find` should be generic and maybe some other methods, but to me this class seems to work more like the `EntityManager` from doctrine/orm rather than a repository.

phpstan.neon.dist
franmomu: Glad to see you back @phansys!

VincentLanglet:
> I understand that `AuditReader::find` should be generic and maybe some other methods, but to me this class seems to work more like the `EntityManager` from doctrine/orm rather than a repository.

For SonataAdmin, it's considered as a repository, and the AuditManager would be the EntityManager
https://github.com/sonata-project/SonataAdminBundle/blob/4.x/src/Model/AuditManagerInterface.php#L42-L46
We can use different AuditReader for different classes.

But I think the generic part of the AuditReader of this project can be removed, 
this shouldn't break the SonataDoctrineORM static analysis, with this code https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/4.x/src/Model/AuditReader.php#L41-L44

phpstan.neon.dist
VincentLanglet: Is there a lot of dataprovider to fix ?

VincentLanglet (dismissed):

VincentLanglet (approved):

src/AuditReader.php
```diff
@@ -774,8 +775,12 @@ protected function getEntityPersister($className)
     /**
      * Simplified and stolen code from UnitOfWork::createEntity.
      *
-     * @param string     $className
-     * @param int|string $revision
+     * @template T of object
+     *
+     * @param string                $className
```
jordisala1991, line 780: isnt this a class-string?

src/AuditReader.php
```diff
@@ -206,9 +203,12 @@ public function clearEntityCache(): void
      * returns last revision INCLUDING "DEL" revision. If you want to throw exception instead, set
      * $threatDeletionAsException to true.
      *
-     * @param string           $className
-     * @param int|string|array $id
-     * @param int|string       $revision
+     * @template T of object
+     *
+     * @param string                                    $className
```
jordisala1991, line 208: isnt this a class-string?

src/AuditReader.php
```diff
@@ -566,15 +566,15 @@ public function findRevisions($className, $id)
      * NEXT_MAJOR: Add NoRevisionFoundException as possible exception.
      * Gets the current revision of the entity with given ID.
      *
-     * @param string           $className
-     * @param int|string|array $id
+     * @param string                               $className
```
jordisala1991, line 569: Another class-string?

src/AuditReader.php
```diff
@@ -508,15 +508,15 @@ public function findRevision($revision)
     /**
      * Find all revisions that were made of entity class with given id.
      *
-     * @param string           $className
-     * @param int|string|array $id
+     * @param string                               $className
```
jordisala1991, line 511: And another one

src/AuditReader.php
franmomu: yep, all these parameters are defined below as `@phpstan-param`

jordisala1991 (approved):

jordisala1991:
Thanks @franmomu

PR #498, opened 2022-06-25
Add rector

jordisala1991:
## Subject

I am targeting this branch, because this is pedantic.


jordisala1991:
The one missing is page bundle but until get release 4.0 we can't add it, since it tries to add type hints everywhere and it is not BC.

VincentLanglet (approved):

PR #508, opened 2022-08-13
Bump to phpstan 8 and psalm 3

VincentLanglet:


src/Action/CompareAction.php
```diff
@@ -37,10 +37,12 @@ public function __invoke(Request $request, string $className, string $id, ?int $
     {
         if (null === $oldRev) {
             $oldRev = $request->query->get('oldRev');
+            \assert(null !== $oldRev);
         }
 
         if (null === $newRev) {
             $newRev = $request->query->get('newRev');
+            \assert(null !== $newRev);
         }
```
jordisala1991, line 46: this code is a bit strange isnt it?

src/Collection/AuditedCollection.php
```diff
@@ -187,7 +187,7 @@ public function containsKey($key)
     }
 
     /**
-     * @return object
+     * @return object|null
      *
      * @phpstan-return T
```
jordisala1991: If you modify the above, this one should be changed too right?

src/Action/CompareAction.php
VincentLanglet: Yes.

The route is
```
simple_things_entity_audit_compare:
    path: /compare/{className}/{id}/{oldRev}/{newRev}
    controller: SimpleThings\EntityAudit\Action\CompareAction
    defaults:
        oldRev: ~
        newRev: ~
```
and the controller
```
    public function compareAction(Request $request, $className, $id, $oldRev = null, $newRev = null)
    {
        $compareAction = new CompareAction($this->get('twig'), $this->getAuditReader());

        return $compareAction($request, $className, $id, $oldRev, $newRev);
    }
```

We might remove the `= null` in next major but I dunno the impacts.

src/Collection/AuditedCollection.php
VincentLanglet: It was an error

src/Action/CompareAction.php
jordisala1991: Well if you are going to compare things, better pasa the things to compare right? I guess we should remove null yes.

VincentLanglet: Sure but remove it so far is a BC break.

Also, I don't understand why it would be in 
```
$request->query->get('oldRev');
```
and not passed by the controller, since the routeParam is `oldRev` too...

Seems like it's coming from https://github.com/sonata-project/EntityAuditBundle/blob/1.x/src/Resources/views/Audit/view_entity.html.twig#L31-L32
Since I wasn't sure about what I could do, I prefer to avoid breaking things

jordisala1991 (approved):

PR #534, opened 2023-02-03
Fix audit revision query for enum column mapping

X-Coder264:
## Subject

I am targeting this branch, because it's a bug fix.

After https://github.com/doctrine/orm/pull/10088 we get an enum object back in the `DEL` revision `saveRevisionEntityData` code path instead of an integer or string value which then blows up on the `$conn->executeStatement()` call with the following exception:

```
Error: Object of class Sonata\EntityAuditBundle\Tests\Fixtures\Issue\Status could not be converted to string

vendor/doctrine/dbal/src/Driver/PDO/Statement.php:56
```

## Changelog

```markdown
### Fixed
- Audit query for Doctrine ORM >= 2.14.1 for entities with enumType column mapping
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "doctrine/common": "^3.1",
         "doctrine/dbal": "^2.13 || ^3.0",
         "doctrine/event-manager": "^1.1",
-        "doctrine/orm": "^2.11",
+        "doctrine/orm": "^2.10",
         "doctrine/persistence": "^2.2 || ^3.0",
         "symfony/config": "^4.4 || ^5.4 || ^6.0",
         "symfony/dependency-injection": "^4.4 || ^5.4 || ^6.0",
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "doctrine/common": "^3.1",
         "doctrine/dbal": "^2.13 || ^3.0",
         "doctrine/event-manager": "^1.1",
-        "doctrine/orm": "^2.10",
+        "doctrine/orm": "^2.11",
         "doctrine/persistence": "^2.2 || ^3.0",
         "symfony/config": "^4.4 || ^5.4 || ^6.0",
         "symfony/dependency-injection": "^4.4 || ^5.4 || ^6.0",
--- a/src/EventListener/LogRevisionsListener.php
+++ b/src/EventListener/LogRevisionsListener.php
@@ -583,7 +583,7 @@ class LogRevisionsListener implements EventSubscriber
         }
 
         foreach ($params as $key => $parameterValue) {
-            if (\interface_exists(\BackedEnum::class) && $parameterValue instanceof \BackedEnum) {
+            if (interface_exists(\BackedEnum::class) && $parameterValue instanceof \BackedEnum) {
                 $params[$key] = $parameterValue->value;
             }
         }
--- a/src/EventListener/LogRevisionsListener.php
+++ b/src/EventListener/LogRevisionsListener.php
@@ -583,7 +583,7 @@ class LogRevisionsListener implements EventSubscriber
         }
 
         foreach ($params as $key => $parameterValue) {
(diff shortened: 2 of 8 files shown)


VincentLanglet (changes requested):
Thanks for the fix, do you have time to take a look at the ci errors ?
You'll need to
- Solve some linters errors
- Solve some "possiblyNull" static analysis error by adding `static::assertNotNull($foo)`.
- Skip the test for PHP version which doesn't support ENUM

src/EventListener/LogRevisionsListener.php
```diff
@@ -582,6 +582,12 @@ private function saveRevisionEntityData(EntityManagerInterface $em, ClassMetadat
             );
         }
 
+        foreach ($params as $key => $parameterValue) {
+            if (interface_exists(\BackedEnum::class) && $parameterValue instanceof \BackedEnum) {
```
phansys: Do we really need the `interface_exists()` check?

composer.json
```diff
@@ -14,7 +14,7 @@
         "doctrine/common": "^3.1",
         "doctrine/dbal": "^2.13 || ^3.0",
         "doctrine/event-manager": "^1.1",
-        "doctrine/orm": "^2.10",
+        "doctrine/orm": "^2.11",
```
phansys: IMO, this change is not related to the described fix.

src/EventListener/LogRevisionsListener.php
X-Coder264: I don't know how PHP behaves when doing an instanceof check against an interface that does not exist, even if that maybe doesn't emit a warning today it doesn't mean it's not going to in the future so I think it's better to keep this as is as it definitively doesn't hurt.

composer.json
X-Coder264: 2.11 added the `enumType` feature for the mapping which is needed to reproduce the issue (and is used in the test) but I see now that you test PHP 8.1 only against `highest` dependencies so I should be able to revert this.

X-Coder264:
Even though the test I've added is marked as requiring PHP 8.1 to run, the PHP 7.4 and PHP 8.0 pipelines are still failing as the entity with the enum mapping class metadata is still being registered (so it blows up in `Doctrine/ORM/Mapping/ClassMetadataFactory`). Does anybody have a suggestion how we can solve this?

VincentLanglet:
> Even though the test I've added is marked as requiring PHP 8.1 to run, the PHP 7.4 and PHP 8.0 pipelines are still failing as the entity with the enum mapping class metadata is still being registered (so it blows up in `Doctrine/ORM/Mapping/ClassMetadataFactory`). Does anybody have a suggestion how we can solve this?

Filtering here maybe https://github.com/sonata-project/EntityAuditBundle/blob/1.x/tests/BaseTest.php#L98-L102

src/EventListener/LogRevisionsListener.php
phansys: PHP handles the absence of the class gracefully. Avoiding superfluous function calls is always desirable.
```suggestion
            if ($parameterValue instanceof \BackedEnum) {
```

X-Coder264: Done

composer.json
X-Coder264: Done

X-Coder264:
> > Even though the test I've added is marked as requiring PHP 8.1 to run, the PHP 7.4 and PHP 8.0 pipelines are still failing as the entity with the enum mapping class metadata is still being registered (so it blows up in `Doctrine/ORM/Mapping/ClassMetadataFactory`). Does anybody have a suggestion how we can solve this?
> 
> Filtering here maybe https://github.com/sonata-project/EntityAuditBundle/blob/1.x/tests/BaseTest.php#L98-L102

@VincentLanglet Thanks - all CI pipelines have now passed :tada:

VincentLanglet (approved):

phansys (approved):

VincentLanglet:
Thanks

PR #536, opened 2023-02-08
Add support for PSR clock

X-Coder264:
## Subject

I am targeting this branch, because it adds a new feature without a BC-break.

## Changelog

```markdown
### Added
- Support for PSR ClockInterface
```

### DIFF
--- a/src/EventListener/LogRevisionsListener.php
+++ b/src/EventListener/LogRevisionsListener.php
@@ -27,6 +27,7 @@ use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Persisters\Entity\EntityPersister;
 use Doctrine\ORM\UnitOfWork;
 use Doctrine\Persistence\Mapping\MappingException;
+use Psr\Clock\ClockInterface;
 use SimpleThings\EntityAudit\AuditConfiguration;
 use SimpleThings\EntityAudit\AuditManager;
 use SimpleThings\EntityAudit\Metadata\MetadataFactory;
@@ -36,6 +37,8 @@ class LogRevisionsListener implements EventSubscriber
     private AuditConfiguration $config;
 
     private MetadataFactory $metadataFactory;
+
+    private ?ClockInterface $clock;
 
     /**
      * @var string[]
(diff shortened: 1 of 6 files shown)


src/EventListener/LogRevisionsListener.php
```diff
@@ -323,15 +327,17 @@ private function getManyToManyRelations(EntityManagerInterface $em, object $enti
...
             $conn->insert(
                 $this->config->getRevisionTableName(),
                 [
-                    'timestamp' => date_create('now'),
+                    'timestamp' => $now,
                     'username' => $this->config->getCurrentUsername(),
                 ],
                 [
-                    Types::DATETIME_MUTABLE,
+                    Types::DATETIME_IMMUTABLE,
```
VincentLanglet, line 340: Isn't this change a BC-break ? I dunno what it imply.

tests/BaseTest.php
```diff
@@ -171,12 +172,22 @@ protected function getAuditManager(): AuditManager
...
         $auditConfig->setUsernameCallable(static fn (): string => 'beberlei');
 
-        $auditManager = new AuditManager($auditConfig);
+        $auditManager = new AuditManager($auditConfig, $this->getClock());
         $auditManager->registerEvents($this->_getConnection()->getEventManager());
 
         return $this->auditManager = $auditManager;
     }
 
+    protected function getClock(): ClockInterface
```
VincentLanglet: This should return null by default instead to test that we correctly work without clock.
```suggestion
    protected function getClock(): ?ClockInterface
    {
         return null;
    }
```

VincentLanglet (changes requested):

tests/BaseTest.php
X-Coder264: Done

src/EventListener/LogRevisionsListener.php
X-Coder264: It's not a BC break. The query either executes or it doesn't (it obviously executes fine as all the tests are passing).

All the types do is tell DBAL which `\Doctrine\DBAL\Types\Type` child class to use to get the value to bind to the query, more specifically in this case whether to call `\Doctrine\DBAL\Types\DateTimeType::convertToDatabaseValue` or `\Doctrine\DBAL\Types\DateImmutableType::convertToDatabaseValue`.

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #509, opened 2022-08-18
Feature/many to many support

pietaj:
I am targeting this branch, because the bundle does not support m:n auditing and the subject was already mentioned 
here https://github.com/sonata-project/EntityAuditBundle/issues/302, and here https://github.com/sonata-project/EntityAuditBundle/pull/321 but was never finished

## Changelog

```markdown
### Added
- Support for ManyToMany
```

Issue #522, opened 2022-10-11
Many To Many audit relation fail with new entity in the relation.


mpoiriert:
#### Sonata packages

<details><summary>show</summary>
<p>

```
$ composer show --latest 'sonata-project/*'
sonata-project/admin-bundle              4.19.0 4.19.0 The missing Symfony Admin Generator
sonata-project/block-bundle              4.18.0 4.18.0 Symfony SonataBlockBundle
sonata-project/cache                     2.2.0  2.2.0  Cache library
Package sonata-project/cache is abandoned, you should avoid using it. No replacement was suggested.
sonata-project/doctrine-extensions       2.0.1  2.0.1  Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 4.8.0  4.8.0  Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/entity-audit-bundle       1.9.0  1.9.0  Audit for Doctrine Entities
sonata-project/exporter                  3.0.0  3.0.0  Lightweight Exporter library
sonata-project/form-extensions           1.18.0 1.18.0 Symfony form extensions
sonata-project/media-bundle              4.5.1  4.5.1  Symfony SonataMediaBundle
sonata-project/translation-bundle        3.1.0  3.1.0  SonataTranslationBundle
sonata-project/twig-extensions           2.0.0  2.0.0  Sonata twig extensions

```

</p>
</details>

#### Symfony packages

<details><summary>show</summary>
<p>

```
$ composer show --latest 'symfony/*'
symfony/amqp-messenger             v5.4.13 v5.4.13 Symfony AMQP extension Messenger Bridge
symfony/asset                      v5.4.13 v6.1.5  Manages URL generation and versioning of web assets such as CSS stylesheets, JavaScript files and image files
symfony/browser-kit                v5.4.11 v6.1.3  Simulates the behavior of a web browser, allowing you to make requests, click on links and submit forms programmatically
symfony/cache                      v5.4.13 v6.1.5  Provides extended PSR-6, PSR-16 (and tags) implementations
symfony/cache-contracts            v2.5.2  v3.1.1  Generic abstractions related to caching
symfony/config                     v5.4.11 v6.1.3  Helps you find, load, combine, autofill and validate configuration values of any kind
symfony/console                    v5.4.13 v6.1.5  Eases the creation of beautiful and testable command line interfaces
[...]

PR #537, opened 2023-02-13
Fix schema creation for self referencing many to many association

X-Coder264:
## Subject

I am targeting this branch, because it's a bug fix.

We've tried to upgrade from 1.7 to 1.9 but our CI was failing with the following exception:

```
Doctrine\DBAL\Schema\Exception\TableDoesNotExist: There is no table with name "foo" in the schema.
/app/vendor/doctrine/dbal/src/Schema/Exception/TableDoesNotExist.php:16
/app/vendor/doctrine/dbal/src/Schema/SchemaException.php:68
/app/vendor/doctrine/dbal/src/Schema/Schema.php:188
/app/vendor/sonata-project/entity-audit-bundle/src/EventListener/CreateSchemaListener.php:102
```

Further debugging showed that the issue was introduced in https://github.com/sonata-project/EntityAuditBundle/pull/509 as the case of a self referencing many to many association which uses class table inheritance was not taken into consideration in the create schema listener class so it currently tries to create the revision/audit table for the related table which does not yet exist at that point. The fix was to execute that logic (if we run into that case) in the post schema generation event instead where we are sure that all schema tables are created so we can then safely create the revision table for them.

## Changelog

```markdown
### Fixed
- Schema creation for self referencing many to many association with class table inheritance
```

### DIFF
--- a/src/EventListener/CreateSchemaListener.php
+++ b/src/EventListener/CreateSchemaListener.php
@@ -32,6 +32,11 @@ class CreateSchemaListener implements EventSubscriber
 
     private MetadataFactory $metadataFactory;
 
+    /**
+     * @var string[]
+     */
+    private array $defferedJoinTablesToCreate = [];
+
     public function __construct(AuditManager $auditManager)
     {
         $this->config = $auditManager->getConfiguration();
@@ -47,6 +52,7 @@ class CreateSchemaListener implements EventSubscriber
     {
         return [
             ToolEvents::postGenerateSchemaTable,
+            ToolEvents::postGenerateSchema,
         ];
     }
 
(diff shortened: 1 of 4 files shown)


VincentLanglet:
Does it solve https://github.com/sonata-project/EntityAuditBundle/issues/522 or is it a different issue ?

X-Coder264:
#522 seems to be about a `NotNullConstraintViolationException` exception that gets thrown while the audit records are being recorded, while this issue is about a `TableDoesNotExist` exception which gets thrown during schema creation (so it blocks any further testing for us of the many to many auditing feature).

VincentLanglet:
Thanks

PR #542, opened 2023-02-20
Update Psalm

X-Coder264:
## Subject

I am targeting this branch, because this is a backwards compatible change.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,8 +36,8 @@
         "phpstan/phpstan-strict-rules": "^1.0",
         "phpstan/phpstan-symfony": "^1.0",
         "phpunit/phpunit": "^9.5.13",
-        "psalm/plugin-phpunit": "^0.17",
-        "psalm/plugin-symfony": "^4.0",
+        "psalm/plugin-phpunit": "^0.18",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -48,8 +48,8 @@
         "symfony/security-bundle": "^4.4 || ^5.4 || ^6.0",
         "symfony/twig-bundle": "^4.4 || ^5.4 || ^6.0",
         "symfony/var-dumper": "^4.4 || ^5.4 || ^6.0",
-        "vimeo/psalm": "^4.8",
-        "weirdan/doctrine-psalm-plugin": "^2.0"
+        "vimeo/psalm": "^4.30 || ^5.7",
```
X-Coder264, line 51: Unfortunately we can't completely drop Psalm v4 as the `psalm/plugin-symfony` v5 package which supports Psalm 5 does not support Symfony 4. When we drop Symfony 4 support then we'll be able to drop Psalm v4 too.

VincentLanglet (approved):

VincentLanglet:
Thanks

VincentLanglet:
It just remember we'll have to try to do the same for every other sonata package 
https://github.com/sonata-project?type=source
🥲

PR #543, opened 2023-02-20
Clear extra updates array to prevent memory leak

X-Coder264:
## Subject

I am targeting this branch, because this is a bug fix.
The `extraUpdates` array property is never being cleared so this results in memory leaks in long running processes (like Symfony Messenger queues) - now it's gonna be cleared on every `$entityManager->clear()` call.

## Changelog

```markdown
### Fixed
- Clear extra updates array to prevent memory leak
```


VincentLanglet (approved):

VincentLanglet:
Thanks

PR #551, opened 2023-03-05
DevKit updates for 1.x branch

SonataCI:
## Changelog

```markdown
### Removed
- Support for PHP 7.4
- Support for Symfony 6.0 and 6.1
```


tests/App/Entity/User.php
```diff
@@ -36,15 +36,13 @@ class User
...
-     */
-    #[ORM\Column(type: Types::STRING)]
-    private string $name;
-
-    public function __construct(string $name)
-    {
-        $this->name = $name;
+    public function __construct(
+        /**
+         * @ORM\Column(type="string")
```
jordisala1991: this can be drop

src/Revision.php
```diff
@@ -18,29 +18,14 @@
...
-    /**
-     * @param int|string $rev
-     */
-    public function __construct($rev, \DateTime $timestamp, ?string $username)
-    {
-        $this->rev = $rev;
-        $this->timestamp = $timestamp;
-        $this->username = $username;
+    public function __construct(
+        private int|string $rev,
```
jordisala1991: I dont think it is safe to do this change

src/AuditReader.php
```diff
@@ -708,7 +690,7 @@ public function getEntityValues($className, $entity)
      * @phpstan-param class-string<T>              $className
      * @phpstan-return array<T|null>
      */
-    public function getEntityHistory($className, $id)
+    public function getEntityHistory($className, int|string|array $id)
```
jordisala1991: I think a lot of this changes are not safe to do in this branch.

VincentLanglet: We have phpstan at a good level, a 86% coverage.
And I kinda consider our BC policy should only be for people using the code according to the phpdoc.

Edit: Oh the class is not final.

jordisala1991: It is not final and we never did a major of this bundle. Not sure if there is a way to avoid disable this rule.

VincentLanglet: I've opened https://github.com/rectorphp/rector/issues/7823 but yes we can disable the rule anyway.

src/AuditReader.php
```diff
@@ -639,7 +633,7 @@ public function getCurrentRevision($className, $id)
      * an object with a given id.
      *
      * @param string     $className
-     * @param int|string $id
+     * @param int|string $revision
```
jordisala1991:
```suggestion
     * @param int|string $id
```

jordisala1991 (approved):

PR #561, opened 2023-03-31
Allow sql keywords as column name

Schrank:
I don't know wether this is a proper PR, but unfortunately I'm overworked but I think this is so important to fix, that maybe someone will fix anything which needs to be done before merging it.

Currently when having a column name like `order` it breaks the audit insert and throws an exception:

    An exception occurred while executing a query: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'order) VALUES (41, 'UPD', '?mjr?L?EBU?m?', '???0,?GI??#?G?', 'Gsadf at line 14

## Subject

I am targeting this branch, because the problem for me lives in 1.12.0.

## Changelog

```markdown
### Security
- Allow SQL keywords in column names by properly escaping them

```

### DIFF
--- a/src/EventListener/LogRevisionsListener.php
+++ b/src/EventListener/LogRevisionsListener.php
@@ -424,7 +424,7 @@ class LogRevisionsListener implements EventSubscriber
                 $placeholders[] = true === ($class->fieldMappings[$field]['requireSQLConversion'] ?? false)
                     ? $type->convertToDatabaseValueSQL('?', $platform)
                     : '?';
-                $sql .= ', `'.$em->getConfiguration()->getQuoteStrategy()->getColumnName($field, $class, $platform).'`';
+                $sql .= ', '.$em->getConfiguration()->getQuoteStrategy()->getColumnName($field, $class, $platform);
             }
 
             if (
--- a/src/EventListener/LogRevisionsListener.php
+++ b/src/EventListener/LogRevisionsListener.php
@@ -384,8 +384,8 @@ class LogRevisionsListener implements EventSubscriber
             $placeholders = ['?', '?'];
             $tableName = $this->config->getTableName($class);
 
-            $sql = 'INSERT INTO '.$tableName.' ('.
-                $this->config->getRevisionFieldName().', '.$this->config->getRevisionTypeFieldName();
+            $sql = 'INSERT INTO '.$tableName.' (`'.
+                $this->config->getRevisionFieldName().'`, `'.$this->config->getRevisionTypeFieldName() . '`';
 
             $fields = [];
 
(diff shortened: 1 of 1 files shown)


VincentLanglet:
Hi @Schrank, thanks for your PR.

Seems like the current tests are failing with your changes, so i'm not sure it works.
Can you take a look ?

Also it would be nice to add a test with a reserved keyword in order to confirm it solve the issue and avoid a regression later.

Schrank:
Some more of my thoughts here, but the TL;DR is below.
<details>
fixed the code style and made the tests green again.

But I'm not sure, wether this is a SQL query which does what it is supposed to do:

INSERT INTO EscapedColumnsEntity_audit (`rev`, `revtype`, id, lft, "left") VALUES (?, ?, ?, ?, ?)

From my point of view, **my PR is superflous**, because my issue is already taken care of, but the escaping shouldn't be `"` but `

My assumption is, that doctrine is taking care of that in `$em->getConfiguration()->getQuoteStrategy()->getColumnName($field, $class, $platform)` but the quoting is wrong.

I dig a little in the code but don't understand the issue. ORDER and LEFT are both part of `\Doctrine\DBAL\Platforms\AbstractMySQLPlatform::getReservedKeywordsClass`

</details>
... Oh stupid me. In case anyone finds this, remember [https://twitter.com/derrabus/status/1609582084670078977](https://twitter.com/derrabus/status/1609582084670078977)

> Use an actual version here, e.g. 8.0.31. Oracle tends to add features in 8.0.x releases these days and DBAL 4 is going to be picky about incomplete server versions.

Talking about doctrine connection string, because if you don't set the version and platform correctly you end up here:

\Doctrine\DBAL\Platforms\AbstractPlatform::getIdentifierQuoteCharacter

instead of here:

\Doctrine\DBAL\Platforms\AbstractMySQLPlatform::getIdentifierQuoteCharacter

Thanks for your time @VincentLanglet!

Schrank:
I'm copying my toot here, because the problem exists and it is NOT about wrong version on the doctrine dsn string. But I don't understand the problem either :-/

I'm fiddeling with https://github.com/sonata-project/EntityAuditBundle/pull/561 [#doctrine](https://phpc.social/tags/doctrine) 

`$em->getConfiguration()->getQuoteStrategy()->getColumnName($field, $class, $platform)`

Can someone tell me, wether the expected behaviour of the return is to quote a columnName called order (mysql keyword) with ` assuming a correct platform like MariaDb1027Platform is set? Because this doesn't happen 😅  Bug or wrong expectation?

Schrank:
Last note on this, [it seems that doctrine doesn't support auto-escaping](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/limitations-and-known-issues.html#identifier-quoting-and-legacy-databases), although I'm not convinced 😅  not sure how the entity insert works and the entity-audit doesn't, but whatever. To make this PR work, I need to make sure it not only works on MySQL and the better solution to this is, to escape the column name on the entity:


As described here: https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/basic-mapping.html#quoting-reserved-words

PR #575, opened 2023-05-08
DevKit updates for 1.x branch

SonataCI:


jordisala1991:
I might need help with this build @VincentLanglet . It is installing non stable version of symfony/string, even with prefer stable.

composer.json
```diff
@@ -53,8 +53,11 @@
     "conflict": {
         "doctrine/doctrine-bundle": "<2.7",
         "gedmo/doctrine-extensions": "<3.7",
-        "symfony/framework-bundle": "<5.4"
+        "symfony/framework-bundle": "<5.4",
+        "symfony/string": "5.4.0-RC1, 5.4.0-BETA1"
```
jordisala1991:
```suggestion
        "symfony/string": "5.4.0-BETA1 5.4.0-RC1"
```

composer.json
```diff
@@ -53,8 +53,11 @@
     "conflict": {
         "doctrine/doctrine-bundle": "<2.7",
         "gedmo/doctrine-extensions": "<3.7",
-        "symfony/framework-bundle": "<5.4"
+        "symfony/framework-bundle": "<5.4",
+        "symfony/string": "5.4.0-BETA1 5.4.0-RC1"
```
jordisala1991:
```suggestion
        "symfony/string": "5.4.0-BETA1 || 5.4.0-RC1"
```

composer.json
```diff
@@ -53,8 +53,11 @@
     "conflict": {
         "doctrine/doctrine-bundle": "<2.7",
         "gedmo/doctrine-extensions": "<3.7",
-        "symfony/framework-bundle": "<5.4"
+        "symfony/framework-bundle": "<5.4",
+        "symfony/string": "5.4.0-BETA1 || 5.4.0-RC1"
```
jordisala1991, line 57: This is how I solved it at the end @VincentLanglet . Composer for some reason prefers to install this rc and beta and keep translation contracts up to v3 instead of lower it down to v2 and use latest 5.4 of symfony string. Seems like a bug in composer but not sure when I will have time for it.

jordisala1991 (approved):

VincentLanglet:
> I might need help with this build @VincentLanglet . It is installing non stable version of symfony/string, even with prefer stable.

Shouldn't be reported as an issue to composer then ?

composer.json
VincentLanglet: Maybe worth to open an issue ?

PR #584, opened 2023-08-02
Do not rely on auto incremental ids

franmomu:


phansys (approved):

VincentLanglet:
Does it mean a BC break should be reported to doctrine/orm ?

VincentLanglet (approved):

PR #587, opened 2023-09-05
Added the with_foreign_keys parameter to allow disabling the creation of foreign keys

SavageDays:
## Subject

Large legacy projects can contain a massive amount of audit data. Adding foreign keys can take a significant amount of time. I have added a 'with_foreign_keys' parameter, which can be used to disable the creation of foreign keys.

## Changelog

```markdown
### Added
Added the `disable_foreign_keys` parameter, which disables the creation of foreign keys.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ simple_things_entity_audit:
     entity_manager: custom
 ```
 
-If there is no need to create foreign keys for any reason, you can specify the corresponding parameter:
+If you need to explicitly discard the foreign keys inferred from the audited entities, you can use the `disable_foreign_keys` parameter:
 ```yaml
 simple_things_entity_audit:
     disable_foreign_keys: true
--- a/README.md
+++ b/README.md
@@ -110,6 +110,12 @@ simple_things_entity_audit:
     entity_manager: custom
 ```
 
+If there is no need to create foreign keys for any reason, you can specify the corresponding parameter:
+```yaml
+simple_things_entity_audit:
+    with_foreign_keys: false
+```
+
 ### Creating new tables
 
 Call the command below to see the new tables in the update schema queue.
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ simple_things_entity_audit:
 If there is no need to create foreign keys for any reason, you can specify the corresponding parameter:
 ```yaml
 simple_things_entity_audit:
-    with_foreign_keys: false
+    disable_foreign_keys: true
 ```
 
 ### Creating new tables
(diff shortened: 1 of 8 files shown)


src/DependencyInjection/Configuration.php
```diff
@@ -54,6 +54,7 @@ public function getConfigTreeBuilder()
                 ->scalarNode('revision_field_name')->defaultValue('rev')->end()
                 ->scalarNode('revision_type_field_name')->defaultValue('revtype')->end()
                 ->scalarNode('revision_table_name')->defaultValue('revisions')->end()
+                ->scalarNode('with_foreign_keys')->defaultValue('default')->end()
```
VincentLanglet: The default value should be a boolean and not `default` no ?

VincentLanglet (commented):
Hi, thanks for the contribution.

Overall the PR seems ok, just:
- I think a `disable_foreign_keys: true` option seems more natural than `with_foreign_keys: false`. WDYT ?
- Maybe the use cases of this option could be in the Readme

src/DependencyInjection/Configuration.php
SavageDays: Yes, I fixed it.

SavageDays: And renamed the parameter name to disable_foreign_keys

README.md
```diff
@@ -110,6 +110,12 @@ simple_things_entity_audit:
     entity_manager: custom
 ```
 
+If there is no need to create foreign keys for any reason, you can specify the corresponding parameter:
```
phansys:
```suggestion
If you need to explicitly discard the foreign keys inferred from the audited entities, you can use the `disable_foreign_keys` parameter:
```
BTW, providing context for the use cases like the example mentioned in the PR's description can be useful here.

src/AuditConfiguration.php
```diff
@@ -79,6 +81,16 @@ public function getTableName(ClassMetadataInfo $metadata)
         return $this->getTablePrefix().$tableName.$this->getTableSuffix();
     }
 
+    public function isDisabledForeignKeys(): bool
```
phansys: Just for semantics, IMO it should be `areForeignKeysDisabled()`.

SavageDays: I have made changes based on both comments.

VincentLanglet (approved):

phansys (approved):

phansys:
Thank you @SavageDays!

PR #599, opened 2024-01-12
Fix "diff" result for dates and other objects

befresh-mweimerskirch:
## Subject

This pull request fixes an issue with the ArrayDiff class and consequentially the AuditReader.

AuditReader::diff currently shows dates (instances of DateTime or DateTimeImmutable) as changed even though they are the same.

This issue was introduced when the loose comparator ("==") was changed to the strict comparison ("===") in commit [bf02fd4](https://github.com/sonata-project/EntityAuditBundle/commit/bf02fd49463c25f23f9d2e7755ec20eec8151733#diff-80e1d2d2d865a9b942c0d8c11dd827f4dcbe88352e95ac38f0079bd416caae68)

This PR reverts the old behaviour for objects, while keeping the strict comparison for other types.

I also added tests. Some of the tests fail without the change, so I hope this illustrates the issue.

I am targeting this branch, because this is a fix that restores previous behaviour (and slightly improves it).

## Changelog

```markdown
### Fixed
- Objects now no longer show as different when their values are the same. This restores some of the old behaviour of the EntityAuditBundle
```

### DIFF
--- a/src/Utils/ArrayDiff.php
+++ b/src/Utils/ArrayDiff.php
@@ -55,8 +55,10 @@ class ArrayDiff
      * Compare the type and the property values of two objects.
      * Return true if they are the same, false otherwise.
      * If the type is the same and all properties are the same, this will return true, even if they are not the same instance.
+     * This method is different from comparing two objects using ==,
+     * because internally the strict comparison operator (===) is used to compare the properties.
      *
-     * @see https://www.php.net/manual/en/language.oop5.object-comparison.php.
+     * @see https://www.php.net/manual/en/language.oop5.object-comparison.php
      */
     private function compareObjects(object $object1, object $object2): bool
     {
--- a/src/Utils/ArrayDiff.php
+++ b/src/Utils/ArrayDiff.php
@@ -37,7 +37,9 @@ class ArrayDiff
             $old = \array_key_exists($field, $oldData) ? $oldData[$field] : null;
             $new = \array_key_exists($field, $newData) ? $newData[$field] : null;
 
-            if ($old === $new) {
+            // If the values are objects, we will compare them by their properties.
+            // This is necessary because the strict comparison operator (===) will return false if the objects are not the same instance.
+            if ((\is_object($old) && \is_object($new) && $this->compareObjects($old, $new)) || ($old === $new)) {
                 $row = ['old' => '', 'new' => '', 'same' => $old];
             } else {
                 $row = ['old' => $old, 'new' => $new, 'same' => ''];
@@ -47,5 +49,40 @@ class ArrayDiff
         }
 
         return $diff;
+    }
+
+    /**
+     * Compare the type and the property values of two objects.
+     * Return true if they are the same, false otherwise.
+     * If the type is the same and all properties are the same, this will return true, even if they are not the same instance.
+     */
+    public function compareObjects(object $object1, object $object2): bool
+    {
+        // Check if the objects are of the same type.
+        $obj1Class = $object1::class;
+        $obj2Class = $object2::class;
+        if ($obj1Class !== $obj2Class) {
+            return false;
+        }
+
+        // Check if all properties are the same.
+        $obj1Properties = (array) $object1;
+        $obj2Properties = (array) $object2;
+        foreach ($obj1Properties as $key => $value) {
+            if (!\array_key_exists($key, $obj2Properties)) {
+                return false;
+            }
+            if (\is_object($value) && \is_object($obj2Properties[$key])) {
+                if (!$this->compareObjects($value, $obj2Properties[$key])) {
+                    return false;
+                }
+                continue;
+            }
+            if ($value !== $obj2Properties[$key]) {
+                return false;
+            }
+        }
+
+        return true;
     }
 }
--- a/src/Utils/ArrayDiff.php
+++ b/src/Utils/ArrayDiff.php
@@ -55,13 +55,13 @@ class ArrayDiff
      * Compare the type and the property values of two objects.
      * Return true if they are the same, false otherwise.
      * If the type is the same and all properties are the same, this will return true, even if they are not the same instance.
+     *
+     * @see https://www.php.net/manual/en/language.oop5.object-comparison.php.
      */
-    public function compareObjects(object $object1, object $object2): bool
+    private function compareObjects(object $object1, object $object2): bool
     {
         // Check if the objects are of the same type.
-        $obj1Class = $object1::class;
-        $obj2Class = $object2::class;
-        if ($obj1Class !== $obj2Class) {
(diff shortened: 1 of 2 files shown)


VincentLanglet (dismissed):

tests/Utils/ArrayDiffTest.php
```diff
@@ -0,0 +1,104 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\EntityAuditBundle\Tests\Utils;
+
+use PHPUnit\Framework\TestCase;
+use SimpleThings\EntityAudit\Utils\ArrayDiff;
+
+class ArrayDiffTest extends TestCase
```
phansys:
```suggestion
final class ArrayDiffTest extends TestCase
```

src/Utils/ArrayDiff.php
```diff
@@ -48,4 +50,39 @@ public function diff($oldData, $newData)
...
+        $obj1Properties = (array) $object1;
+        $obj2Properties = (array) $object2;
+        foreach ($obj1Properties as $key => $value) {
+            if (!\array_key_exists($key, $obj2Properties)) {
+                return false;
+            }
+            if (\is_object($value) && \is_object($obj2Properties[$key])) {
+                if (!$this->compareObjects($value, $obj2Properties[$key])) {
+                    return false;
+                }
```
phansys, line 80:
```suggestion
                }

```

src/Utils/ArrayDiff.php
```diff
@@ -48,4 +50,39 @@ public function diff($oldData, $newData)
...
+     * Compare the type and the property values of two objects.
+     * Return true if they are the same, false otherwise.
+     * If the type is the same and all properties are the same, this will return true, even if they are not the same instance.
+     */
+    public function compareObjects(object $object1, object $object2): bool
+    {
+        // Check if the objects are of the same type.
+        $obj1Class = $object1::class;
+        $obj2Class = $object2::class;
+        if ($obj1Class !== $obj2Class) {
```
phansys:
```suggestion
        if ($object1::class !== $object2::class) {
```
The only concern I have regarding this is about class aliases, but I don't know if that is valid use case for this method.

tests/Utils/ArrayDiffTest.php
```diff
@@ -0,0 +1,104 @@
...
+        $result = $diff->diff($array1, $array2);
+
+        static::assertSame($expected, $result);
+    }
+
+    public function testDiffDateDifferent(): void
+    {
+        $diff = new ArrayDiff();
+
+        $dateInstance1 = new \DateTimeImmutable('2014-01-01 00:00:00.000000');
```
phansys, line 65: IMO, we should also add a test setting different time zones in the compared objects, in order to expose how the comparison behaves with different time zones using the same UTC offset.

src/Utils/ArrayDiff.php
```diff
@@ -48,4 +50,39 @@ public function diff($oldData, $newData)
 
         return $diff;
     }
+
+    /**
+     * Compare the type and the property values of two objects.
+     * Return true if they are the same, false otherwise.
+     * If the type is the same and all properties are the same, this will return true, even if they are not the same instance.
```
phansys, line 57:
```suggestion
     * If the type is the same and all properties are the same, this will return true, even if they are not the same instance.
     *
     * @see https://www.php.net/manual/en/language.oop5.object-comparison.php.
```

src/Utils/ArrayDiff.php
```diff
@@ -48,4 +50,39 @@ public function diff($oldData, $newData)
...
+     */
+    public function compareObjects(object $object1, object $object2): bool
```
phansys:
```suggestion
    private function compareObjects(object $object1, object $object2): bool
```

phansys (changes requested):
Thank you @befresh-mweimerskirch.

Could you please update the description telling how the method `compareObjects()` is different from the [loose comparison (`==`) between objects](https://www.php.net/manual/en/language.oop5.object-comparison.php)?

VincentLanglet:
> Thank you @befresh-mweimerskirch.
> 
> Could you please update the description telling how the method `compareObjects()` is different from the [loose comparison (`==`) between objects](https://www.php.net/manual/en/language.oop5.object-comparison.php)?

If I understood correctly, the compareObject is pretty the same than `==` for object.
That's just our phpstan-strict-rule which forbid using loose comparison.

src/Utils/ArrayDiff.php
befresh-mweimerskirch: I think class aliases are out of scope of how the EntityAuditBundle is used.

tests/Utils/ArrayDiffTest.php
befresh-mweimerskirch: I'll add that test case on Monday.

befresh-mweimerskirch:
@phansys 
> Could you please update the description telling how the method `compareObjects()` is different from the [loose comparison (`==`) between objects](https://www.php.net/manual/en/language.oop5.object-comparison.php)?

Added to the todo section of the PR. I'll do that on Monday.

@VincentLanglet 
> If I understood correctly, the compareObject is pretty the same than `==` for object. That's just our phpstan-strict-rule which forbid using loose comparison.

No. When using the loose comparison operator on objects, the property values are also compared loosely. Imagine a class that has only a property "name". One object with name "TEST" and one object with name "test" as their respective value. Those two instances would be considered equal using the "==" operator.
My method on the other hand uses strict comparison for the values.

src/Utils/ArrayDiff.php
```diff
@@ -48,4 +50,40 @@ public function diff($oldData, $newData)
 
         return $diff;
     }
+
+    /**
+     * Compare the type and the property values of two objects.
+     * Return true if they are the same, false otherwise.
+     * If the type is the same and all properties are the same, this will return true, even if they are not the same instance.
+     *
+     * @see https://www.php.net/manual/en/language.oop5.object-comparison.php.
```
VincentLanglet:
```suggestion
     * @see https://www.php.net/manual/en/language.oop5.object-comparison.php
```
(The link with the dot is a 404 for me)

befresh-mweimerskirch:
As promised, I added a test case with two dates that have different timezones.
And I described how the method compareObjects() is different from the loose comparison (==) between objects

src/Utils/ArrayDiff.php
befresh-mweimerskirch: Fixed as part of my previous commit

VincentLanglet (approved):

phansys (approved):

VincentLanglet:
Thanks

PR #621, opened 2024-04-22
Replace EntityManager type-hint with EntityManagerInterface

EmmanuelVella:
## Subject

This PR fixes an issue when using an entity manager decorator (same as https://github.com/sonata-project/EntityAuditBundle/issues/149).

I am targeting this branch, because it's BC (EntityManagerInterface was introduced in doctrine orm `2.4`, and version `^2.7` is required in `composer.json`.

## Changelog

```markdown
### Fixed
- Replaced AuditManager::createAuditReader(EntityManager $em) type-hint with EntityManagerInterface

```

### DIFF
--- a/src/AuditManager.php
+++ b/src/AuditManager.php
@@ -14,7 +14,7 @@ declare(strict_types=1);
 namespace SimpleThings\EntityAudit;
 
 use Doctrine\Common\EventManager;
-use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityManagerInterface;
 use Psr\Clock\ClockInterface;
 use SimpleThings\EntityAudit\EventListener\CreateSchemaListener;
 use SimpleThings\EntityAudit\EventListener\LogRevisionsListener;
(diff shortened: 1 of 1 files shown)


src/AuditManager.php
```diff
@@ -52,11 +52,9 @@ public function getConfiguration()
     }
 
     /**
-     * NEXT_MAJOR: Use `\Doctrine\ORM\EntityManagerInterface` for argument 1.
-     *
      * @return AuditReader
      */
-    public function createAuditReader(EntityManager $em)
+    public function createAuditReader(EntityManagerInterface $em)
```
dmaicher, line 57: This is a BC break however. Classes which are extending `AuditReader` and overwriting `createAuditReader` will break due to this change.

Hence we planned to do this in the next major release as the comment above suggested.

EmmanuelVella: Sure I understand. I'm closing the PR. Thank you !

PR #631, opened 2024-07-25
allow ManyToMany associations with referenced column name different from 'id'

awm1:
## Subject

I've found out that if I want to audit entity with ID column with name different from 'id' :

class User
{
    /**
     * @ORM\Id()
     * @ORM\Column(name="user_name",type="string")
     */
    private $userName;
    /**
     * @ORM\ManyToMany(targetEntity="AppBundle\Entity\Department", inversedBy="members", cascade={"persist"})
     * @ORM\JoinTable(
     *     name="systemusers_department_membership",
     *     joinColumns={@ORM\JoinColumn(name="systemusers_id", referencedColumnName="user_name")},
     *     inverseJoinColumns={@ORM\JoinColumn(name="department_id", referencedColumnName="id")}
     * )
     */
    private $departmentMembership;
}

...those ManyToMany assoc isn't referred by its correct identifier name, but static 'id' string which will throw Exception. So I try to get it from available source, however I'm not sure if this is the right way to obtain it. I am targeting 1.x branch because this is simple fix which should not break anything.

## Changelog

```markdown
### Fixed
Accessing ManyToMany associations of audited entity with identity field of name different from 'id'.  


```


src/AuditReader.php
```diff
@@ -1022,7 +1022,7 @@ private function createEntity($className, array $columnMap, array $data, $revisi
                     foreach (self::getRelationToSourceKeyColumns($assoc) as $sourceKeyJoinColumn => $sourceKeyColumn) {
                         $whereId[] = "{$sourceKeyJoinColumn} = ?";
 
-                        $reflField = $classMetadata->reflFields['id'];
+                        $reflField = $classMetadata->reflFields[$classMetadata->getFieldName($sourceKeyColumn) ?? 'id'];
```
phansys: If I'm not wrong, I guess we could use something like this:
```suggestion
                        $reflField = $classMetadata->getSingleIdReflectionProperty();
```

phansys (commented):
Thank you @awm1!

Please, add a test covering this fix.

src/AuditReader.php
awm1: Yes, you are right, this is much better approach and it seems to work in the same way.

.gitignore
```diff
@@ -2,6 +2,7 @@
 #
 # It's auto-generated by sonata-project/dev-kit package.
 
+.idea/
```
phansys, line 5: This change is not related to the PR.

PR #637, opened 2024-12-02
doctrine/dbal 4 support

ghost:
## Subject

Support for doctrine/dbal 4.

Closes #636.

## Changelog

```markdown
### Added
- doctrine/dbal 4.x support
```

PR #652, opened 2025-07-24
Add support for doctrine/dbal 4

### LINKED ISSUE #636: Support for doctrine/dbal 4
## Feature Request

The latest version of sonata-project/entity-audit-bundle, 1.18.0, requires doctrine/dbal ^3.6. This precludes its use in projects that use doctrine/dbal 4, first released in February of 2024.

A list of BC-breaking changes are available in the [doctrine/dbal 4.0 upgrade guide](https://github.com/doctrine/dbal/blob/4.0.0/UPGRADE.md).

If there are not significant BC breaks involving DBAL features that EntityAuditBundle uses, support for both DBAL 3 and 4 could be released in a version 1.19.0.

If maintaining support for both DBAL major versions simultaneously would prove difficult, a separate 2.0.0 version that drops support for doctrine/dbal 3 would make more sense.

valkars:
## Subject
I am targeting this branch, because this patch is backwards compatible.

While adding support for Doctrine DBAL 4, I encountered a bug during testing: when creating audit tables for @JoinTable associations, the length parameter is not respected if the primary key is a string. In my case, the country table has a primary key of type CHAR(2). With DBAL 3, the audit table was created with a column of type VARCHAR(255). In DBAL 4, however, an exception is thrown because string columns now require an explicit length. I’ve fixed this issue.
I then tested this branch on my project using both DBAL 3 and DBAL 4, and everything is working as expected.

Closes #636.

## Changelog
```markdown
### Added
- Support for doctrine/dbal 4

### Fixed
- Creating of Join Audit tables with string primary keys
```

### DIFF
--- a/src/EventListener/CreateSchemaListener.php
+++ b/src/EventListener/CreateSchemaListener.php
@@ -15,8 +15,12 @@ namespace SimpleThings\EntityAudit\EventListener;
 
 use Doctrine\Common\EventSubscriber;
 use Doctrine\DBAL\Schema\Column;
+use Doctrine\DBAL\Schema\Name\Identifier;
+use Doctrine\DBAL\Schema\Name\UnqualifiedName;
+use Doctrine\DBAL\Schema\PrimaryKeyConstraint;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\Types\StringType;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\DBAL\Types\Types;
 use Doctrine\ORM\Mapping\ClassMetadata;
@@ -26,6 +30,7 @@ use Doctrine\ORM\Tools\ToolEvents;
 use SimpleThings\EntityAudit\AuditConfiguration;
 use SimpleThings\EntityAudit\AuditManager;
 use SimpleThings\EntityAudit\Metadata\MetadataFactory;
+use SimpleThings\EntityAudit\Utils\DbalCompatibilityTrait;
 use SimpleThings\EntityAudit\Utils\ORMCompatibilityTrait;
 
 /**
@@ -36,6 +41,7 @@ use SimpleThings\EntityAudit\Utils\ORMCompatibilityTrait;
  */
 class CreateSchemaListener implements EventSubscriber
 {
+    use DbalCompatibilityTrait;
     use ORMCompatibilityTrait;
 
     private AuditConfiguration $config;
@@ -95,9 +101,13 @@ class CreateSchemaListener implements EventSubscriber
         $revisionsTable = $this->createRevisionsTable($schema);
 
         $entityTable = $eventArgs->getClassTable();
-        $revisionTable = $schema->createTable(
-            $this->config->getTablePrefix().$entityTable->getName().$this->config->getTableSuffix()
-        );
+        if ($this->isDbal4_3()) {
+            $tableName = $this->config->getTablePrefix().$entityTable->getObjectName()->toString().$this->config->getTableSuffix();
+        } else {
+            /** @psalm-suppress InternalMethod */
+            $tableName = $this->config->getTablePrefix().$entityTable->getName().$this->config->getTableSuffix(); // @phpstan-ignore-line
+        }
+        $revisionTable = $schema->createTable($tableName);
 
         foreach ($entityTable->getColumns() as $column) {
             $this->addColumnToTable($column, $revisionTable);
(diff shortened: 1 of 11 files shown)


valkars:
@VincentLanglet when you have some time, could you please review my pull request?

src/EventListener/CreateSchemaListener.php
```diff
@@ -143,12 +145,29 @@ public function postGenerateSchema(GenerateSchemaEventArgs $eventArgs): void
...
+        $revisionForeignKeyName = $this->config->getRevisionFieldName().'_'.\md5($relatedTable->getName()).'_fk';
         $primaryKey = $revisionsTable->getPrimaryKey();
         \assert(null !== $primaryKey);
 
+        /**
+         * doctrine/dbal 3 support - Table::addForeignKeyConstraint() takes a Table instead of a string
+         *
+         * NEXT_MAJOR: remove this `if` block
+         */
+        if (\version_compare(InstalledVersions::getVersion('doctrine/dbal') ?? '', '4.0.0', '<')) {
```
VincentLanglet: Let's try to avoid using `InstalledVersions::getVersion` ; I don't think we ever used it on sonata/project.

What can be done, is something as `class_exists(Foo::class)` (or `!class_exists(Bar::class)` when Foo was added in 4.0 or Bar removed in 4.0.

src/EventListener/LogRevisionsListener.php
```diff
@@ -186,7 +188,11 @@ public function postFlush(PostFlushEventArgs $eventArgs): void
                     $types[] = self::getMappingValue($meta->fieldMappings[$idField], 'type');
                 } elseif (isset($meta->associationMappings[$idField]['joinColumns'])) {
                     $columnName = self::getMappingNameValue($meta->associationMappings[$idField]['joinColumns'][0]);
-                    $types[] = $meta->associationMappings[$idField]['type'];
+                    if (\version_compare(InstalledVersions::getVersion('doctrine/dbal') ?? '', '4.0.0', '<')) {
```
VincentLanglet: Same here (Maybe checking for class_exists(ParameterType::class) ? I dunno if it was introduced in 4.0)

src/EventListener/LogRevisionsListener.php
```diff
@@ -383,8 +389,15 @@ private function getRevisionId(Connection $conn)
...
 
             $revisionId = $conn->lastInsertId();
-            if (false === $revisionId) {
-                throw new \RuntimeException('Unable to retrieve the last revision id.');
+            /**
+             * Preceding lastInsertId throws Doctrine\DBAL\Exception\DriverException on doctrine/dbal 4+, making the next check unnecessary.
+             *
+             * NEXT_MAJOR: Remove the following block
+             */
+            if (\version_compare(InstalledVersions::getVersion('doctrine/dbal') ?? '', '4.0.0', '<')) {
```
VincentLanglet: No need for this if.

The check will just be usefull for 3.x.

You can just write
```
if (false === $revisionId) { // @phpstan-ignore-line doctrine/dbal 3 lastInsertId() can return false
                    throw new \RuntimeException('Unable to retrieve the last revision id.');
                }
```

src/EventListener/LogRevisionsListener.php
```diff
@@ -523,7 +536,11 @@ private function saveRevisionEntityData(EntityManagerInterface $em, ClassMetadat
         $conn = $em->getConnection();
 
         $params = [$this->getRevisionId($conn), $revType];
-        $types = [\PDO::PARAM_INT, \PDO::PARAM_STR];
+        if (\version_compare(InstalledVersions::getVersion('doctrine/dbal') ?? '', '4.0.0', '<')) {
```
VincentLanglet: Same here (Maybe checking for class_exists(ParameterType::class) ? I dunno if it was introduced in 4.0)

src/EventListener/LogRevisionsListener.php
```diff
@@ -547,7 +564,11 @@ private function saveRevisionEntityData(EntityManagerInterface $em, ClassMetadat
                         $fields[$sourceColumn] = true;
                         if (null === $data) {
                             $params[] = null;
-                            $types[] = \PDO::PARAM_STR;
+                            if (\version_compare(InstalledVersions::getVersion('doctrine/dbal') ?? '', '4.0.0', '<')) {
```
VincentLanglet: Same here

src/EventListener/LogRevisionsListener.php
```diff
@@ -642,7 +663,11 @@ private function recordRevisionForManyToManyEntity(
     ): void {
         $conn = $em->getConnection();
         $joinTableParams = [$this->getRevisionId($conn), $revType];
-        $joinTableTypes = [\PDO::PARAM_INT, \PDO::PARAM_STR];
+        if (\version_compare(InstalledVersions::getVersion('doctrine/dbal') ?? '', '4.0.0', '<')) {
```
VincentLanglet: Same here

VincentLanglet (commented):
There is lot of failure for tests

src/EventListener/LogRevisionsListener.php
valkars: Ok, I'll try to find new class in Dbal 4 and add check for existence

src/EventListener/CreateSchemaListener.php
```diff
@@ -95,8 +101,9 @@ public function postGenerateSchemaTable(GenerateSchemaTableEventArgs $eventArgs)
         $revisionsTable = $this->createRevisionsTable($schema);
 
         $entityTable = $eventArgs->getClassTable();
+        /** @psalm-suppress InternalMethod */
```
VincentLanglet: Is there a way to avoid the use of internal methods ?

If not, should we open an issue on doctrine side to remove the `@internal` ?

src/EventListener/CreateSchemaListener.php
```diff
@@ -105,15 +112,29 @@ public function postGenerateSchemaTable(GenerateSchemaTableEventArgs $eventArgs)
...
-        $pkColumns[] = $this->config->getRevisionFieldName();
-        $revisionTable->setPrimaryKey($pkColumns);
-        $revIndexName = $this->config->getRevisionFieldName().'_'.md5($revisionTable->getName()).'_idx';
+        if ($this->isDbal4_3()) {
+            $primaryKey = $entityTable->getPrimaryKeyConstraint();
+            \assert(null !== $primaryKey);
+            $pkColumns = $primaryKey->getColumnNames();
+            /** @psalm-suppress ArgumentTypeCoercion */
+            $pkColumns[] = new UnqualifiedName(Identifier::unquoted($this->config->getRevisionFieldName())); // @phpstan-ignore-line
+            /** @psalm-suppress InternalMethod */
```
VincentLanglet: Is there a way to avoid the use of internal methods ?

If not, should we open an issue on doctrine side to remove the `@internal` ?

src/EventListener/CreateSchemaListener.php
```diff
@@ -105,15 +112,29 @@ public function postGenerateSchemaTable(GenerateSchemaTableEventArgs $eventArgs)
...
+            /** @psalm-suppress DeprecatedMethod */
+            $primaryKey = $entityTable->getPrimaryKey();
+            \assert(null !== $primaryKey);
+            /** @psalm-suppress DeprecatedMethod */
+            $pkColumns = $primaryKey->getColumns();
+            $pkColumns[] = $this->config->getRevisionFieldName();
+            /** @psalm-suppress DeprecatedMethod */
+            $revisionTable->setPrimaryKey($pkColumns);
+        }
+        /** @psalm-suppress InternalMethod */
```
VincentLanglet: Is there a way to avoid the use of internal methods ?

If not, should we open an issue on doctrine side to remove the `@internal` ?

src/EventListener/CreateSchemaListener.php
```diff
@@ -143,25 +164,55 @@ public function postGenerateSchema(GenerateSchemaEventArgs $eventArgs): void
...
+        }
     }
 
     /**
      * Copies $column to another table. All its options are copied but notnull and autoincrement which are set to false.
      */
     private function addColumnToTable(Column $column, Table $targetTable): void
     {
-        $columnName = $column->getName();
+        /** @psalm-suppress InternalMethod */
```
VincentLanglet: Is there a way to avoid the use of internal methods ?

If not, should we open an issue on doctrine side to remove the `@internal` ?

VincentLanglet (dismissed):
Awesome

valkars:
These methods became internal in Dbal 4.3, so for this version I've removed usage of them

src/EventListener/CreateSchemaListener.php
valkars: Fixed for Dbal 4.3 (made internal)

VincentLanglet (approved):

VincentLanglet:
Awesome

VincentLanglet:
I'll wait few days before the release.
If you want you can target 1.x on your project to confirm it works fine :)

valkars:
Ok, I'll try
This PR is not needed anymore https://github.com/sonata-project/EntityAuditBundle/pull/637

valkars:
2 days live in our project, all good @VincentLanglet

PR #683, opened 2026-06-27
Fix audit schema generation for reserved-keyword table names on DBAL 4.3+

gulaandrij:
## Subject

On `doctrine/dbal >= 4.3`, audit schema generation crashes with `Doctrine\DBAL\Schema\Exception\InvalidState: Object name has not been initialized` for any audited entity whose table name is an SQL reserved keyword (and is therefore force-quoted).

`CreateSchemaListener` derives the audit table name from `$entityTable->getObjectName()->toString()`, which on DBAL >= 4.3 renders a **quoted** identifier for reserved keywords (e.g. `"user"`). Concatenating it with the configured table prefix/suffix yields `"user"_audit`, which is not a parseable identifier: DBAL silently leaves the new table's name uninitialized and then throws from `getObjectName()` on the next access (here, while building the revision index name).

The pre-4.3 code path used `Table::getName()` (which returns the *unquoted* name), so it was not affected.

This PR adds `DbalCompatibilityTrait::getUnquotedTableName()`, which on DBAL >= 4.3 returns the raw unquoted identifier value (`getObjectName()->getUnqualifiedName()->getValue()`, prefixed with the qualifier when present) and otherwise falls back to `Table::getName()`. It is used at the three audit-table-name derivation sites (the entity table and the many-to-many join tables), restoring the historical behaviour.

A regression test (`IssueReservedTableNameTest`, with an entity mapped to a reserved `` `user` `` table) reproduces the crash before the fix and passes after it. The full test suite, PHPStan, PHP-CS-Fixer and Rector all pass on `doctrine/dbal` 4.4.3.

I am targeting the `1.x` branch, because this is a backwards-compatible bug fix.

## Changelog

```markdown
### Fixed
- Crash during audit schema generation for entities whose table name is a reserved SQL keyword on `doctrine/dbal` 4.3+
```

### DIFF
--- a/src/EventListener/CreateSchemaListener.php
+++ b/src/EventListener/CreateSchemaListener.php
@@ -98,11 +98,7 @@ class CreateSchemaListener implements EventSubscriber
         $revisionsTable = $this->createRevisionsTable($schema);
 
         $entityTable = $eventArgs->getClassTable();
-        if ($this->isDbal4_3()) {
-            $tableName = $this->config->getTablePrefix().$entityTable->getObjectName()->toString().$this->config->getTableSuffix();
-        } else {
-            $tableName = $this->config->getTablePrefix().$entityTable->getName().$this->config->getTableSuffix(); // @phpstan-ignore-line
-        }
+        $tableName = $this->config->getTablePrefix().$this->getUnquotedTableName($entityTable).$this->config->getTableSuffix();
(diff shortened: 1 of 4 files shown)


gulaandrij:
@VincentLanglet ping

VincentLanglet (approved):

VincentLanglet:
Lot of CI checks are failing

